PR #17, opened 2026-08-01
chore: release hygiene — dated-CHANGELOG gate, drop the executed plan doc

shivamlalakiya:
The tag/version/changelog gate in `publish.yml` matched `^## \[$V\]`, which `## [0.7.0] - TBD` satisfies.

0.7.0 and 1.0.0 are both merged to main and deliberately staged with `- TBD` until their release window. Under the old gate, tagging either today would have shipped a CHANGELOG whose release date reads **TBD** — and the gate would have waved it through.

Now requires `## [X.Y.Z] - YYYY-MM-DD` and prints what it actually found, so the failure explains itself.

Verified against the current CHANGELOG:

| Version | old gate | new gate |
|---|---|---|
| 0.6.0 (dated) | pass | pass |
| 0.7.0 (TBD) | **pass** | fail |
| 1.0.0 (TBD) | **pass** | fail |


shivamlalakiya:
Also folded in: **removed `HARDENING_PLAN.md`.** Every W1–W4 step is implemented and merged (0.6.0 on PyPI; 0.7.0 and 1.0.0 on main awaiting their windows). It was a planning artifact rather than documentation, nothing references it, and it never shipped in a distribution — it was only tracked because `git add -A` swept it into the 0.6.0 commit. Recoverable from history at `cc32428`.

Issue #23, opened 2026-08-02
Cover the two documented transform-time guards that no test exercises (MatchingGiftFeaturizer, ShareOfWalletScorer)


shivamlalakiya:
## Summary

Two input-validation guards are documented in public docstrings but never executed by the test suite. Both are transform-time twins of guards that *are* tested at fit time. This issue adds the two missing tests. **No source code changes** — both guards already behave correctly.

This is a good first issue: pure input validation, no fundraising domain knowledge needed, and the whole change is roughly eight lines across two test files.

## The gap

**1. `MatchingGiftFeaturizer.transform` — non-DataFrame input**

`philanthropy/preprocessing/_matching_gift.py:186`

```python
if not isinstance(X, pd.DataFrame):
    raise TypeError("X must be a pandas DataFrame")
```

This is advertised in the `transform` docstring (`Raises / TypeError / If X is not a pandas DataFrame`, `_matching_gift.py:177-182`). The identical guard in `fit` (`_matching_gift.py:147`) *is* tested — `tests/test_matching_gift.py:132-135` — but nothing ever calls `transform` with a non-DataFrame, so line 186 never runs.

**2. `ShareOfWalletScorer.transform` — out-of-range `capacity_col_idx`**

`philanthropy/preprocessing/_share_of_wallet.py:479`

```python
if cap_idx >= X_arr.shape[1]:
    raise ValueError(
        f"`capacity_col_idx` ({cap_idx}) exceeds number of columns "
        f"({X_arr.shape[1]})."
    )
```

`fit` only checks `capacity_col_idx >= 0` (tested at `tests/test_share_of_wallet.py:351-353`). The upper bound is checked in `transform` only. That means this succeeds today:

```python
ShareOfWalletScorer(capacity_col_idx=5).fit(np.ones((6, 3)))   # OK
```

and only fails later, at `transform`. That deferred failure is exactly the error a user hits mid-pipeline, so it is worth pinning the message text.

## Steps

1. **Set up.**
   ```bash
   git clone https://github.com/PhilanthroPy-Project/PhilanthroPy
   cd PhilanthroPy
   python -m pip install -e ".[dev]"
   ```
   (Editable install matters — a non-editable copy in site-packages will shadow your working tree under pytest.)

2. **Confirm the gap for yourself.**
   ```bash
   python -m pytest tests/ --cov=philanthropy --cov-report=term-missing -q
   ```
[...]

PR #30, opened 2026-08-03
Fix tautological doctest and add no-leakage test for default gap_years=0

### LINKED ISSUE #26: Replace the tautological doctest in FiscalYearGroupedSplitter and add a real no-leakage test for the default gap_years=0
## Summary

`philanthropy/model_selection/_temporal_donor_splitter.py:32` contains a doctest that can never fail. It runs in CI on every push and verifies nothing — on the one class whose entire stated purpose is "guarantees zero data leakage across fiscal years".

This is a good first issue: one line of source to fix, one small test to add, no API design decisions, no domain knowledge beyond the rule already written in the module docstring.

## The bug

`philanthropy/model_selection/_temporal_donor_splitter.py:31-33`:

```python
>>> train_idx, test_idx = splits[0]
>>> bool(fiscal_years[test_idx].max() <= fiscal_years[train_idx].min() + 1 or True)
True
```

The trailing `or True` short-circuits the comparison away. Convince yourself in one command, no clone needed:

```console
$ python -c "print(bool(0 <= -999 + 1 or True))"
True
```

Doctests are a real CI gate — `.github/workflows/ci.yml:66`:

```yaml
- name: Docstring examples
  run: python -m pytest philanthropy --doctest-modules -q --no-cov
```

So this is a green check that would stay green even if `split()` started handing future fiscal years to the training fold. It is also user-facing: that expression is what someone reads in the rendered docs when they want to know what the splitter guarantees.

## Scope note (so nobody is misled)

The **class** docstring at lines 90-94 does already assert the real property (`assert train_fy.max() < test_fy.min(), "No leakage"`) for `gap_years=0`, and that one does run. So the guarantee is not completely unchecked today. Two things are still worth fixing:

1. the tautology on line 32 is simply wrong and should not ship, and
2. in `tests/`, the only leakage assertion is `test_gap_years_withholds_the_year_before_each_test_fold` (`tests/test_model_selection.py:63-74`), which covers `gap_years=1` only. The **default** `gap_years=0` — the config every user gets from `FiscalYearGroupedSplitter(n_splits=3)` — has its guarantee living in a docstring rather than in the test suite, which is not where a reviewer looks for it.

## Steps

1. **Set up**

   ```bash
   git clone https://github.com/PhilanthroPy-Project/PhilanthroPy
   cd PhilanthroPy
   ```

   The editable install matters — a non-editable copy in `site-packages` will shadow your edits under pytest.

2. **Fix the doctest.** In `philanthropy/model_selection/_temporal_donor_splitter.py`, keep line 31 (`>>> train_idx, test_idx = splits[0]`) and line 33 (`True`) exactly as they are. Replace line 32 with:

   ```
   >>> bool(fiscal_years[train_idx].max() < fiscal_years[test_idx].min())
   ```

   Please don't touch lines 23-30 or any code outside the module docstring — a one-line diff here is the ideal.

3. **Check it:**

   ```bash
   python -m pytest philanthropy/model_selection/_temporal_donor_splitter.py --doctest-modules -q --no-cov
   ```

   Expect `2 passed` (module docstring + class docstring).

4. **Add one test** to `tests/test_model_selection.py`, right after `test_gap_years_withholds_the_year_before_each_test_fold` (it ends at line 74). Reuse the module-level `_FY_GROUPS` already defined at line 60 — no new fixtures needed. For the **default** splitter `FiscalYearGroupedSplitter(n_splits=3)` (i.e. `gap_years=0`), assert for every `(train_idx, test_idx)` that:

   - `fy[train_idx].max() < fy[test_idx].min()` — no future fiscal year ever appears in training, and
   - `len(set(fy[test_idx])) == 1` — each test fold is exactly one fiscal year.

   Optional bonus, and it does hold: the documented expanding window, i.e. `len(train_idx)` strictly increases across consecutive splits (you'll see 4, 6, 8).

   Match the style of the tests around it — plain `def test_...():`, plain `assert`, no parametrize needed.

5. **Check it:**

   ```bash
   python -m pytest tests/test_model_selection.py -q --no-cov
   ```

   Expect `11 passed` (up from 10).

## Acceptance

```bash
python -m pytest philanthropy/model_selection/_temporal_donor_splitter.py --doctest-modules -q --no-cov \
  && python -m pytest tests/test_model_selection.py -q --no-cov \
  && ! grep -q 'or True' philanthropy/model_selection/_temporal_donor_splitter.py \
  && echo OK
```

`2 passed`, then `11 passed`, then `OK`.

## Running the full gate

The two pytest commands in the acceptance block above are the fast check for this PR. `make ci` is green on `main` and takes about three minutes; running it before you push is welcome but not required for a change this size.

## Domain knowledge needed

Light. The only rule you need is the one already stated at the top of the module: a training fold may never contain a fiscal year at or after the test fold's fiscal year.

## Estimated time

About 30-45 minutes, most of it the initial `pip install -e ".[dev]"`.

---

Want it? Comment "claiming this" and it's yours — happy to answer questions in the thread or on the PR, and no need to have contributed before. If you get stuck at any step, say where and I'll unblock you.

fuleinist:
## Summary

Fixes the tautological doctest in `_temporal_donor_splitter.py` (line 32 used `or True`, making the assertion impossible to fail) and adds a proper no-leakage test for the default `gap_years=0` config — which previously had its guarantee only in a docstring, not the test suite.

Fixes #26

## Changes

**Doctest fix** (1 line):
```python
# Before (tautology — always True regardless of data):
>>> bool(fiscal_years[test_idx].max() <= fiscal_years[train_idx].min() + 1 or True)

# After (real property — fails if leakage occurs):
>>> bool(fiscal_years[train_idx].max() < fiscal_years[test_idx].min())
```

**New test**: `test_default_splitter_no_leakage_gap_years_zero` — covers `gap_years=0` (the default, which every user gets) and asserts:
- `fy[train_idx].max() < fy[test_idx].min()` — no future fiscal year in training
- `len(set(fy[test_idx])) == 1` — each test fold is exactly one fiscal year
- Expanding window: `train_sizes` strictly increases across splits

## Verification

```
$ python -m pytest philanthropy/model_selection/_temporal_donor_splitter.py --doctest-modules -q --no-cov
2 passed

$ python -m pytest tests/test_model_selection.py -q --no-cov
11 passed  (was 10)

$ ! grep -q 'or True' philanthropy/model_selection/_temporal_donor_splitter.py && echo OK
OK
```

### DIFF
--- a/philanthropy/model_selection/_temporal_donor_splitter.py
+++ b/philanthropy/model_selection/_temporal_donor_splitter.py
@@ -29,7 +29,7 @@ Typical usage
 >>> len(splits)
 3
 >>> train_idx, test_idx = splits[0]
->>> bool(fiscal_years[test_idx].max() <= fiscal_years[train_idx].min() + 1 or True)
+>>> bool(fiscal_years[train_idx].max() < fiscal_years[test_idx].min())
 True
 """
 
--- a/tests/test_model_selection.py
+++ b/tests/test_model_selection.py
@@ -74,6 +74,28 @@ def test_gap_years_withholds_the_year_before_each_test_fold():
         assert fy[train_idx].max() < test_fy - 1
 
 
+def test_default_splitter_no_leakage_gap_years_zero():
+    """Default gap_years=0: training fold never contains a FY at or after the test FY,
+    and each test fold is exactly one fiscal year."""
+    X = np.zeros((10, 2))
+    splitter = FiscalYearGroupedSplitter(n_splits=3)
+    splits = list(splitter.split(X, groups=_FY_GROUPS))
+    assert len(splits) == 3
+
+    fy = np.asarray(_FY_GROUPS)
+    train_sizes = []
+    for train_idx, test_idx in splits:
+        # No leakage: max training FY is strictly less than min test FY.
+        assert fy[train_idx].max() < fy[test_idx].min()
+        # Each test fold is exactly one fiscal year.
+        assert len(set(fy[test_idx])) == 1
+        train_sizes.append(len(train_idx))
+
+    # Expanding window: training set grows with each split.
+    assert train_sizes == sorted(train_sizes)
+    assert len(set(train_sizes)) == len(train_sizes)  # strictly increasing
+
+
 def test_not_enough_fiscal_years_names_the_shortfall():
     X = np.zeros((10, 2))
     with pytest.raises(


shivamlalakiya (approved):
Approved — thank you, and welcome.

Exactly the change #26 asked for: one-line source diff plus one test, nothing else touched. Verified locally — acceptance block passes, and I mutation-tested the new assertions to confirm they actually fail when leakage is introduced (the whole point of the issue).

Two cosmetic nits, not worth a round-trip, merging as-is:

1. The strictly-increasing check is two asserts where `assert all(a < b for a, b in zip(train_sizes, train_sizes[1:]))` does it in one.
2. Neighbouring tests use inline comments rather than a docstring.

Merging once the test matrix finishes. The CI delay was on my side — GitHub holds fork workflows from first-time contributors until a maintainer approves the run. Nothing you did wrong.

shivamlalakiya:
@fuleinist  Following up now that this is merged: you're credited in [CONTRIBUTORS.md](https://github.com/PhilanthroPy-Project/PhilanthroPy/blob/main/CONTRIBUTORS.md) and in the `[Unreleased]` changelog (#35) — being the first person other than me in that file is not a small thing.

If you want a second one, the closest match to what you just did is **#31**: 11 `get_feature_names_out` overrides in `philanthropy.preprocessing` render as blank signatures on the docs site. Same shape as #26 — the issue lists every file and line, the acceptance check is one command, and the real work is reading each `transform` so the docstring names the actual columns instead of a generic sentence. ~45–60 min.

Adjacent, if you'd rather: **#32** and **#33** are the same job on the models and the three transformers. **#23** is closer to this PR — two documented transform-time guards that no test exercises.

Comment "claiming this" on whichever and it's yours. Happy to answer questions in the thread.

shivamlalakiya:
@fuleinist thanks again for #30. You caught a doctest that was asserting against itself, which is the kind of bug a test suite is structurally blind to: it passes, it stays green forever, and it never once checks the thing it claims to check. The no-leakage test you added for the default `gap_years=0` is still the one guarding that path.

fuleinist:
@shivamlalakiya thanks for the merge, and for the careful read.

Worth stating up front: I was using an automated PR loop, so this is "setup experience" as logged by automation, not a human's late-night frustration. Still a data point for #188.

Setup itself: nothing tripped. Clone, install, tests, push, PR, all along the path the README lays out. Issue #26 carried the real weight: exact file, exact line, and a repro you can run with one `python -c` line, no clone needed. For an automated contributor that's the ideal first-issue shape, and I'd hold it up as a template.

The annoying thing about #30 was the bug, not the setup. `bool(... or True)` is a test that passes on every push and verifies nothing. The suite stayed green while the splitter's one promise, zero leakage across fiscal years, was guarded by a module docstring and nothing else. The doctest line was cosmetic; the no-leakage test is the real change.

One nit for the discussion: nothing in the code said whether the `+1` in `<= train.min() + 1` was a deliberate fence or an off-by-one. The rule lived only in the docstring. A one-line comment on that comparison would have saved a couple of minutes of head-scratching, and it's the kind of thing that ages well.

PR #40, opened 2026-08-06
docs: document fit and transform on preprocessing transformers

### LINKED ISSUE #33: docs: CRMCleaner, FiscalYearTransformer and WealthPercentileTransformer document neither fit nor transform
## Summary

`CRMCleaner`, `FiscalYearTransformer`, and `WealthPercentileTransformer` have no docstring on `fit` or `transform` — 6 methods. The API reference is generated by mkdocstrings from a bare `::: philanthropy.preprocessing` directive (`docs/reference/preprocessing.md:3`), so all six render on the docs site as a signature with nothing under it.

These three are the only transformers in the package whose `fit`/`transform` pair is undocumented; every other transformer already documents both.

This is a good first issue: no logic changes, no API decisions, and there is a fully documented sibling in the repo to copy the shape from.

## Reproduce

```console
$ python -c "
import inspect, philanthropy.preprocessing as P
print([f'{c}.{n}' for c in ('CRMCleaner','FiscalYearTransformer','WealthPercentileTransformer')
       for n, f in inspect.getmembers(getattr(P, c), inspect.isfunction)
       if n in ('fit','transform')
       and f.__qualname__.split('.')[0] == c
       and not (f.__doc__ or '').strip()])
"
```

Prints all 6 today.

## The 6 methods

| File | Lines | Class |
|---|---|---|
| `philanthropy/preprocessing/_transformers.py` | 91, 108 | `CRMCleaner` |
| `philanthropy/preprocessing/_transformers.py` | 169, 183 | `FiscalYearTransformer` |
| `philanthropy/preprocessing/_wealth_percentile.py` | 22, 52 | `WealthPercentileTransformer` |

## What to write

NumPy format (`mkdocs.yml:64` sets `docstring_style: numpy`). Copy the shape from `WealthScreeningImputer.fit`/`transform` in `philanthropy/preprocessing/_wealth.py` (`transform` is at line 252) — summary line, `Parameters`, `Returns`, `Raises`.

The one thing that matters more than the boilerplate: **say what state `fit` freezes, because that is this package's core contract.** From `CONTRIBUTING.md` and `CLAUDE.md` — all fitted statistics are computed from training data in `fit` and frozen before `transform`, and `transform` is idempotent. Concretely:

- `WealthPercentileTransformer.fit` (`_wealth_percentile.py:22`) freezes `feature_names_in_`, `imputed_cols_`, and `percentile_lookup_` — the last being the **sorted training values per wealth column**, which is the reference distribution `transform` ranks against. A reader needs to know that percentiles are relative to the training cohort, not to the batch being transformed. That is the leakage-safety guarantee in one sentence.
- `CRMCleaner.fit` (`_transformers.py:91`) and `FiscalYearTransformer.fit` (`:169`) fit **no** statistics — they validate `fiscal_year_start` and the input, then return `self`. Say so plainly; "this transformer is stateless" is useful information, not a gap.

Also worth one line each: both `CRMCleaner.transform` and `FiscalYearTransformer.transform` return `np.ndarray | pd.DataFrame` — mention that the return type mirrors the input type, since the signature alone does not tell a reader which they get.

Add a `Raises` section where the method calls `check_is_fitted`, and note the `ValueError("Complex data not supported")` path where it exists.

Do not change any code, only add docstrings.

## Steps

1. **Set up**

   ```bash
   git clone https://github.com/PhilanthroPy-Project/PhilanthroPy
   cd PhilanthroPy
   ```

   The editable install matters — a non-editable copy in `site-packages` will shadow your edits.

2. Add a docstring to each of the 6 methods in the table.

3. **Check it:**

   ```bash
   python -c "
   import inspect, sys, philanthropy.preprocessing as P
   m = [f'{c}.{n}' for c in ('CRMCleaner','FiscalYearTransformer','WealthPercentileTransformer')
        for n, f in inspect.getmembers(getattr(P, c), inspect.isfunction)
        if n in ('fit','transform')
        and f.__qualname__.split('.')[0] == c
        and not (f.__doc__ or '').strip()]
   print('MISSING:', m or 'none'); sys.exit(1 if m else 0)
   " && echo OK
   ```

4. Confirm nothing else moved:

   ```bash
   python -m pytest tests/test_preprocessing.py tests/test_leakage.py -q --no-cov
   ```

## Acceptance

Step 3 prints `MISSING: none` then `OK`, and both commands in step 4 pass.

## Domain knowledge needed

Light, and it is all written down. Read the "Leakage-safety contract" section of `CLAUDE.md` and skim `tests/test_leakage.py` for what the package promises. Describing that promise accurately for `WealthPercentileTransformer` is the substantive part of this issue.

## Estimated time

About 45 minutes.

## Scope note

`get_feature_names_out` on these same three classes, and the undocumented methods in `philanthropy.models`, are filed separately so more than one person can work in parallel. If you touch `_transformers.py` here and someone else is in the same file for the other issue, the conflicts are trivial — different methods.

---

Want it? Comment "claiming this" and it's yours. Happy to answer questions in the thread or on the PR, and no need to have contributed before. If you get stuck at any step, say where and I'll unblock you.

slegarraga:
Closes #33.

Adds NumPy-style docstrings to `fit` and `transform` on:
- `CRMCleaner`
- `FiscalYearTransformer`
- `WealthPercentileTransformer`

The docstrings describe frozen state, return types, and raised errors. `WealthPercentileTransformer` documents that percentiles are relative to the frozen training distribution in `percentile_lookup_`.

Verified:
- missing-docstring check prints `MISSING: none` then `OK`
- `pytest --doctest-modules philanthropy` passes (27)
- `pytest tests/test_preprocessing.py tests/test_leakage.py` passes (49)

### DIFF
--- a/philanthropy/preprocessing/_transformers.py
+++ b/philanthropy/preprocessing/_transformers.py
@@ -89,6 +89,25 @@ class CRMCleaner(TransformerMixin, BaseEstimator):
         self.fiscal_year_start = fiscal_year_start
 
     def fit(self, X, y=None) -> "CRMCleaner":
+        """Validate configuration and input without learning state.
+
+        Parameters
+        ----------
+        X : array-like of shape (n_samples, n_features)
+            Training-set feature matrix.
+        y : ignored
+            Present for scikit-learn API compatibility.
+
+        Returns
+        -------
+        self : CRMCleaner
+            Fitted transformer. This transformer is stateless.
+
+        Raises
+        ------
+        ValueError
+            If ``fiscal_year_start`` is invalid or ``X`` contains complex data.
+        """
         validate_fiscal_year_start(self.fiscal_year_start)
         
         # Try standard validation, fallback to object for mixed-type DataFrames or promotion errors
@@ -106,6 +125,25 @@ class CRMCleaner(TransformerMixin, BaseEstimator):
         return self
 
     def transform(self, X) -> np.ndarray | pd.DataFrame:
+        """Clean CRM dates and amounts using the fitted column configuration.
+
+        Parameters
+        ----------
+        X : array-like of shape (n_samples, n_features)
+            Feature matrix (training or held-out).
+
+        Returns
+        -------
+        X_out : np.ndarray or pd.DataFrame
+            Cleaned feature matrix. The return type mirrors the input type.
+
+        Raises
+        ------
+        sklearn.exceptions.NotFittedError
+            If :meth:`fit` has not been called yet.
+        ValueError
+            If ``X`` contains complex data.
+        """
         check_is_fitted(self)
         try:
             X_arr = validate_data(self, X, dtype=None, ensure_all_finite="allow-nan", reset=False)
@@ -167,6 +205,25 @@ class FiscalYearTransformer(TransformerMixin, BaseEstimator):
         self.fiscal_year_start = fiscal_year_start
 
     def fit(self, X, y=None) -> "FiscalYearTransformer":
+        """Validate configuration and input without learning state.
+
+        Parameters
+        ----------
+        X : array-like of shape (n_samples, n_features)
+            Training-set feature matrix.
+        y : ignored
+            Present for scikit-learn API compatibility.
+
+        Returns
+        -------
+        self : FiscalYearTransformer
+            Fitted transformer. This transformer is stateless.
+
+        Raises
+        ------
+        ValueError
+            If ``fiscal_year_start`` is invalid or ``X`` contains complex data.
+        """
         validate_fiscal_year_start(self.fiscal_year_start)
         try:
             X_validated = validate_data(self, X, dtype=None, ensure_all_finite="allow-nan", reset=True)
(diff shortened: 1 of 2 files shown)


shivamlalakiya:
Thanks for this — the parameter/return/raises coverage on these four methods is exactly the shape the rest of the codebase uses, and calling out that `percentile_lookup_` freezes the *training* distribution is the right thing to make explicit.

One correction before this goes in. Both `CRMCleaner.transform` and `FiscalYearTransformer.transform` document the return as:

> The return type mirrors the input type.

That isn't what the code does. The branch at `philanthropy/preprocessing/_transformers.py:128` (and :215) is:

```python
if _get_pandas_output(self):
    return X_df
return X_df.to_numpy()
```

`_get_pandas_output` reads the scikit-learn `set_output` config, not the type of `X`. With the default config, passing a DataFrame in still gets an ndarray back — so the docstring would send readers the wrong way on the one thing they'd look this up for.

Suggested wording:

> X_out : np.ndarray or pd.DataFrame
>     Cleaned feature matrix. Returns a DataFrame when the transformer is
>     configured with `set_output(transform="pandas")`, otherwise an ndarray.

Fix those two spots and I'll merge. Everything else in the PR is accurate as written.

slegarraga:
Fixed in b9f9b3d. Both `CRMCleaner.transform` and `FiscalYearTransformer.transform` now document that the return is a DataFrame only when `set_output(transform="pandas")` is configured, otherwise an ndarray. Tests pass (49). Thank you for the precise review!

PR #65, opened 2026-08-11
docs: add Raises section to five metrics functions

### LINKED ISSUE #61: docs: five `philanthropy.metrics` functions document Returns but never their Raises
## Summary

Five functions across `philanthropy/metrics/` fully document `Parameters` and
`Returns` but have no `Raises` section, even though each one explicitly raises
`ValueError` on bad input:

| File | Function | Raises today |
|---|---|---|
| `philanthropy/metrics/_financial.py` | `donor_lifetime_value` | `ValueError` for negative `retention_rate`, `lifespan_years`, or `discount_rate` |
| `philanthropy/metrics/_concentration.py` | `gift_concentration_gini` | `ValueError` (via `_clean_nonneg_amounts`) on negative amounts |
| `philanthropy/metrics/_concentration.py` | `top_donor_share` | `ValueError` on negative amounts, and on `top_fraction` outside `(0.0, 1.0]` |
| `philanthropy/metrics/_fairness.py` | `selection_rate_by_group` | `ValueError` on length mismatch, empty `y_pred`, or NaN group labels |
| `philanthropy/metrics/_fairness.py` | `disparate_impact_ratio` | `ValueError` propagated from `selection_rate_by_group` |

`gift_concentration_gini` and `top_donor_share` allude to the negative-amounts
case in the `Parameters` prose ("negatives raise"), but none of the five has a
formal `Raises` section a reader can scan for, and none names the exception
type explicitly.

### Where

| File | Locate with |
|---|---|
| `philanthropy/metrics/_financial.py` | `grep -n "^def donor_lifetime_value" philanthropy/metrics/_financial.py` |
| `philanthropy/metrics/_concentration.py` | `grep -n "^def " philanthropy/metrics/_concentration.py` |
| `philanthropy/metrics/_fairness.py` | `grep -n "^def " philanthropy/metrics/_fairness.py` |

### What to change

NumPy format. Add a `Raises` section to each of the five functions above,
naming `ValueError` and, in one line, the condition that triggers it (the
table above lists the conditions already — no need to re-derive them). For
`disparate_impact_ratio`, it's enough to say the error is propagated from
`selection_rate_by_group` rather than re-explaining the three conditions.

### Tests to add or extend

None — doc-only change; the `ValueError` paths themselves are already covered
by `tests/test_financial.py`, `tests/test_concentration.py`, and
`tests/test_fairness.py` (confirmed via `pytest --cov-report=term-missing`:
all three files show 100% coverage on their respective source modules).

### Done when

```bash
python -c "
import sys, philanthropy.metrics as M
targets = ['donor_lifetime_value', 'gift_concentration_gini', 'top_donor_share',
           'selection_rate_by_group', 'disparate_impact_ratio']
missing = [n for n in targets if 'Raises' not in (getattr(M, n).__doc__ or '')]
print('MISSING:', missing or 'none'); sys.exit(1 if missing else 0)
" && echo OK
```

shubhrai23:
Resolves #61

**What changed:** Added explicit `Raises` sections to the docstrings of five metrics functions (`donor_lifetime_value`, `gift_concentration_gini`, `top_donor_share`, `selection_rate_by_group`, `disparate_impact_ratio`).
**Why:** To document the exact conditions under which these functions raise `ValueError`s, ensuring that the documented contract matches the explicit runtime validations.

### DIFF
--- a/philanthropy/metrics/_concentration.py
+++ b/philanthropy/metrics/_concentration.py
@@ -42,6 +42,11 @@ def gift_concentration_gini(amounts: Collection) -> float:
     float
         Gini coefficient in ``[0.0, 1.0]``. Returns ``0.0`` for an empty input
         or an all-zero total (no distribution to measure).
+
+    Raises
+    ------
+    ValueError
+        If any gift amounts are negative.
     """
     a = _clean_nonneg_amounts(amounts)
     total = a.sum()
@@ -70,6 +75,11 @@ def top_donor_share(amounts: Collection, top_fraction: float = 0.1) -> float:
     float
         Share in ``[0.0, 1.0]``. Returns ``0.0`` for an empty input or an
         all-zero total.
+
+    Raises
+    ------
+    ValueError
+        If ``top_fraction`` is not in ``(0.0, 1.0]``, or if any gift amounts are negative.
     """
     if not 0.0 < top_fraction <= 1.0:
         raise ValueError("top_fraction must be in (0.0, 1.0].")
--- a/philanthropy/metrics/_fairness.py
+++ b/philanthropy/metrics/_fairness.py
@@ -40,6 +40,11 @@ def selection_rate_by_group(
     -------
     dict
         Mapping of group value -> selection rate in ``[0.0, 1.0]``.
+
+    Raises
+    ------
+    ValueError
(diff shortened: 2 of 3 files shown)


shivamlalakiya (approved):
Verified against source — Raises sections are accurate. Thanks, approving.

PR #66, opened 2026-08-11
test: add predict_affinity_score fallback coverage

### LINKED ISSUE #57: test: `DonorPropensityModel.predict_affinity_score` never exercises its single-column decision_function fallback
### What's wrong

`DonorPropensityModel.predict_affinity_score` has three branches on the shape of
`decision_function`'s output: 1-D (binary), 2-D with more than one column
(multiclass), and 2-D with exactly one column (the `df.ravel()` fallback). No
test in `tests/test_donor_propensity_model.py` or `tests/test_propensity.py`
produces a `decision_function` output with exactly one column, so that fallback
line has never run.

### Where

| File | Locate with |
|---|---|
| `philanthropy/models/_propensity.py` | `grep -n "Multiclass case" -A 4 philanthropy/models/_propensity.py` |

### What to change

1. Open `tests/test_donor_propensity_model.py` and find the existing
   `predict_affinity_score` tests (locate with
   `grep -n "def test.*affinity_score" tests/test_donor_propensity_model.py`).
2. Construct (or fake via a fitted estimator whose `decision_function` you
   monkeypatch to return a `(n_samples, 1)` array) the one-column case and
   assert `predict_affinity_score` returns `np.round(df.ravel() * 100, 2)`
   rather than raising or mis-indexing.

### Tests to add or extend

- File: `tests/test_donor_propensity_model.py`
- Add: `test_predict_affinity_score_single_column_decision_function`

### Done when

```bash
python -m pytest tests/test_donor_propensity_model.py tests/test_propensity.py \
  --cov=philanthropy.models._propensity --cov-report=term-missing -q
```

`philanthropy/models/_propensity.py`'s `Missing` column no longer lists the
`df.ravel()` line inside `DonorPropensityModel.predict_affinity_score`.

shubhrai23:
Resolves #57

**What changed:** Added unit test coverage for the single-class fallback branch in `predict_affinity_score`.
**Why:** To ensure that the fallback logic behaves correctly when only one class is present in the training data, improving overall test reliability and coverage.

### DIFF
--- a/tests/test_donor_propensity_model.py
+++ b/tests/test_donor_propensity_model.py
@@ -189,6 +189,25 @@ def test_affinity_score_is_float_array(fitted_model, donor_Xy):
     assert np.issubdtype(scores.dtype, np.floating)
 
 
+def test_predict_affinity_score_single_class(donor_Xy):
+    """Test predict_affinity_score when fit on a single class."""
+    X, y = donor_Xy
+    
+    # Test all zeros
+    y_zeros = np.zeros_like(y)
+    model0 = DonorPropensityModel(n_estimators=5, random_state=0)
+    model0.fit(X, y_zeros)
+    scores0 = model0.predict_affinity_score(X)
+    assert np.all(scores0 == 0.0)
+
+    # Test all ones
+    y_ones = np.ones_like(y)
+    model1 = DonorPropensityModel(n_estimators=5, random_state=0)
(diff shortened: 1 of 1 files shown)


shivamlalakiya (approved):
Verified against source — traced the single-class fallback branch in decision_function/predict_affinity_score, test values (0.0 / 100.0) match actual behavior. Thanks, approving.

PR #69, opened 2026-08-13
fix: LapsePredictor silently scores column-reordered DataFrames

shivamlalakiya:
## Summary
- `LapsePredictor` was the only estimator validating input with `check_array`/`check_X_y` instead of `validate_data` (the convention every sibling estimator follows, per `AGENTS.md`).
- Consequence: it never set `feature_names_in_`, so passing a DataFrame with reordered columns at predict time was silently scored instead of raising — every other estimator (e.g. `DonorPropensityModel`) raises `ValueError` on the same input.

## Why
Found during a 2026-08-13 audit pass. Reproduced before fixing:
```
LapsePredictor.feature_names_in_ set: False
predict(X[["c","b","a"]]) -> no error, wrong predictions
DonorPropensityModel.predict(X[["c","b","a"]]) -> ValueError (correct)
```
A gift-officer call list built from a differently-ordered export would be silently wrong with no signal.

## Changes
- `philanthropy/models/_lapse.py`: swap `check_X_y`/`check_array` for `validate_data` in `fit`/`predict`/`predict_proba`, matching `_wallet.py`. Dropped the now-redundant re-validation inside `predict_lapse_score` since it calls `predict_proba`, which already validates.
- `tests/test_audit_regressions.py`: new regression test that fits on a DataFrame and asserts reordered-column input raises on `predict`/`predict_proba`/`predict_lapse_score`.
- `CHANGELOG.md`: entry under `[Unreleased] / Fixed`.

## Test plan
- [x] `make ci` — exit 0 (lint, mypy, doctests, full suite, 92% coverage gate)
- [x] `make riskcov` — exit 0, 95% against the 93% risk-tier floor
- [x] New regression test fails against pre-fix code, passes against this diff

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,10 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   preferred disclosure channel.
 
 ### Fixed
+- `LapsePredictor` validated input with `check_array`/`check_X_y` instead of
+  `validate_data`, the convention every other estimator follows. It never set
+  `feature_names_in_`, so a DataFrame with reordered columns was silently
+  scored instead of raising — the only estimator in the package with that gap.
 - `mkdocs.yml` had no `site_url`, so the generated `sitemap.xml` was empty and all
   38 documentation pages were uncrawlable, with no `rel=canonical` anywhere.
 - `CONTRIBUTING.md` documented a risk-tier coverage command measuring `metrics/` and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,10 +54,12 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   preferred disclosure channel.
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


shivamlalakiya:
Pushed a follow-up commit (2082c2d) that closes the same defect in `experimental.UpliftTLearner`.

Fixing only the estimator the audit named left the sibling broken. `UpliftTLearner` was the other estimator validating with `check_X_y`/`check_array` instead of `validate_data`, so it never recorded `feature_names_in_` either, and `predict_uplift_score` on a column-reordered DataFrame returned a plausible wrong answer instead of raising. A grep for `check_X_y|check_array` across `philanthropy/` now returns nothing, so the class is closed rather than one instance of it.

Also corrected this branch's own CHANGELOG entry, which asserted LapsePredictor was "the only estimator in the package with that gap". That was false as written; it now names both and says both carry a regression test.

Verified on the branch: `make ci` exit 0, 1690 passed, 2 skipped, coverage 95.50% against the 92% floor, `make riskcov` 95% against the 93% floor.

PR #71, opened 2026-08-13
fix: MatchingGiftFeaturizer had zero check_estimator coverage

shivamlalakiya:
## Summary
- `MatchingGiftFeaturizer` set `tags._skip_test = True`, which skips the *entire* `check_estimator` battery (1 check instead of 46) instead of excluding it from `_STANDARD_ESTIMATORS` with a documented reason — the pattern `RFMTransformer` already uses one section up in the same file.
- It genuinely can't run the generic battery (it hard-requires two named DataFrame columns, one of them categorical text, so it raises `TypeError` on the bare numeric ndarrays the battery feeds), but that needs a written reason, not a silent skip.
- Consequence: the unqualified "every public estimator passes `check_estimator`" claim in README, `paper.md`, and two docs pages was false. This is the kind of thing a JOSS reviewer or a sklearn-ecosystem reader actually verifies.
- Separately found while auditing this: `FinancialForecastModel` had the identical gap — never added to `_STANDARD_ESTIMATORS`, no reason on record — except it actually passes the full battery cleanly. Added.

## Why
Found during a 2026-08-13 audit. Reproduced before fixing:
```python
check_estimator(MatchingGiftFeaturizer())  # with _skip_test forced off
# -> TypeError: X must be a pandas DataFrame
```
```python
check_estimator(FinancialForecastModel(max_iter=20, random_state=0))
# -> passes cleanly, just never added to the list
```

## Changes
- `philanthropy/preprocessing/_matching_gift.py`: remove the `tags._skip_test = True` line (it was pure dead weight — the class was never in `_STANDARD_ESTIMATORS` for it to skip anything from).
- `tests/test_sklearn_compliance.py`:
  - Add `FinancialForecastModel(max_iter=20, random_state=0)` to `_STANDARD_ESTIMATORS`.
  - Add a "SECTION 1c" comment block documenting *why* `MatchingGiftFeaturizer` is excluded, mirroring the existing `RFMTransformer` section, pointing at its existing manual coverage in `tests/test_matching_gift.py` and `tests/test_transformer_leakage_guards.py`.
  - Add `_MANUALLY_COVERED` (a reasoned exemption registry, same shape as `test_public_api_contract.py`'s `_EXEMPT`) and two new tests: `test_every_exemption_carries_a_reason` and `test_every_public_estimator_is_covered_by_the_battery_or_documented`, which cross-references `philanthropy.models.__all__` / `philanthropy.preprocessing.__all__` against the battery + the exemption registry. This is the regression guard — a new public estimator landing with silent `check_estimator` coverage can't happen again without this test failing.
- `README.md`, `paper.md`, `docs/explanation/security_review_answers.md`, `docs/explanation/design_principles.md`, `docs/index.md`: corrected the "every estimator passes `check_estimator`" claim to name the one real, already-documented exception (`UpliftTLearner`, whose `fit(X, y, treatment)` signature breaks the contract by design).
- `CHANGELOG.md`: entry under `[Unreleased] / Fixed`.

## Test plan
- [x] `rtk proxy make ci` — 1741 passed, 2 skipped, 95.50% coverage (gate 92%)
- [x] `rtk proxy make riskcov` — 95% (floor 93%)
- [x] `mkdocs build --strict` — exit 0, no new warnings/errors
- [x] Targeted: `pytest tests/test_sklearn_compliance.py tests/test_matching_gift.py tests/test_transformer_leakage_guards.py tests/test_forecast_model.py tests/test_public_api_contract.py tests/test_doc_examples.py tests/test_examples.py` — 1218 passed, 2 skipped
- [x] New invariant test (`test_every_public_estimator_is_covered_by_the_battery_or_documented`) passes now; verified it fails if `MatchingGiftFeaturizer` is removed from `_MANUALLY_COVERED` without being added back to the battery


shivamlalakiya:
One spot this misses: `AGENTS.md:5` still carries the same claim.

> Every estimator is pipeline-safe, leakage-safe, and passes `sklearn.utils.estimator_checks.check_estimator`.

This PR corrects the claim in `README.md`, `paper.md`, `docs/index.md`, `docs/explanation/design_principles.md` and `docs/explanation/security_review_answers.md`, but `AGENTS.md` is not in the diff, so main will still assert it after this merges.

It matters a bit more than a normal docs miss, because `AGENTS.md` is what `CLAUDE.md` includes, so it is the sentence every agent working in this repo reads as ground truth before touching an estimator. Worth folding into this PR rather than a follow-up, since it is the same claim.

Separately, and worth a thought rather than a change here: the `SECTION 1d` invariant you added walks `models.__all__` and `preprocessing.__all__`. It does not walk `experimental.__all__`, where `UpliftTLearner` lives, and that class is genuinely not `check_estimator` compliant (its `fit` takes an extra `treatment` argument, documented in its own docstring). So the invariant cannot currently catch a new non-compliant estimator landing in `experimental`. Adding that module to the sweep with an explicit exemption entry for `UpliftTLearner` would close it, but it is your call whether `experimental` should be held to the invariant at all.

PR #74, opened 2026-08-16
Add test coverage for constituent_events_to_features guard branches

### LINKED ISSUE #51: test: two guards in constituent_events_to_features go untested — empty-after-filter and missing sourceSystem
### What's wrong

`constituent_events_to_features` has two small guards that no test exercises:

1. After dropping events whose `createdAt` can't be parsed to a timestamp, the function checks `if df.empty` and returns the typed empty frame. The existing "unparseable timestamp" test only drops *some* rows (one bad event alongside one good one), so the frame is never empty at that check — it's never reached with a non-trivial input. Note this is a **different** code path from the "call with `[]`" test, which returns before any of this filtering runs.
2. When the `sourceSystem` column is entirely absent from the input, `distinct_source_systems` defaults to `0`. The existing `test_distinct_source_systems` test only covers the column being *present*.

### Where

| File | Locate with |
|---|---|
| `philanthropy/ingest/_constituent_events.py` | `grep -n "return _empty_feature_frame()" philanthropy/ingest/_constituent_events.py` |
| `philanthropy/ingest/_constituent_events.py` | `grep -n "distinct_source_systems.*= 0" philanthropy/ingest/_constituent_events.py` |

### What to change

1. Open `tests/test_ingest.py` and find `test_unparseable_timestamp_row_dropped` (locate with `grep -n "def test_unparseable_timestamp_row_dropped" tests/test_ingest.py`). Add a sibling test right after it where *every* event has an unparseable `createdAt`, and assert the result is the typed empty frame (same shape assertions as `test_empty_input_returns_typed_empty_frame`).
2. Find `test_distinct_source_systems` (locate with `grep -n "def test_distinct_source_systems" tests/test_ingest.py`). The `_event` helper always sets `sourceSystem` (default `"GIVECAMPUS"`), so it can't produce a missing column — build the event as a plain dict instead, the way `test_missing_event_type_does_not_crash` builds its dict without `eventType`. Add a sibling test right after `test_distinct_source_systems` using one or more dicts with `constituentEmail`/`createdAt`/`amount` but no `sourceSystem` key at all, and assert `distinct_source_systems == 0`.

### Tests to add or extend

- File: `tests/test_ingest.py`
- Add: `test_all_unparseable_timestamps_returns_empty_frame`
- Add: `test_distinct_source_systems_defaults_to_zero_when_column_absent`

### Done when

```bash
python -m pytest tests/test_ingest.py --cov=philanthropy.ingest._constituent_events -q
python -m coverage report --include='philanthropy/ingest/_constituent_events.py' --fail-under=100
make ci
make riskcov
```

The `coverage report --fail-under=100` line exits non-zero until `_constituent_events.py` reads fully covered; `make ci` and `make riskcov` must also stay green.

### First time here?

Read [CONTRIBUTING.md](../../CONTRIBUTING.md) and [AGENTS.md](../../AGENTS.md). Add a `## [Unreleased]` CHANGELOG entry and yourself to `CONTRIBUTORS.md` in the same PR.

AsavariCharati:
Resolves #51

Adds the two missing tests for `constituent_events_to_features`:

- `test_all_unparseable_timestamps_returns_empty_frame` — every event has an unparseable `createdAt`, so all rows are dropped by the `_ts.notna()` filter, leaving `df` empty and hitting the second `_empty_feature_frame()` return (distinct from the empty-input path, which returns before any filtering runs). Same shape assertions as `test_empty_input_returns_typed_empty_frame`.
- `test_distinct_source_systems_defaults_to_zero_when_column_absent` — builds events as plain dicts (not the `_event()` helper, which always sets `sourceSystem`) with the key fully absent, asserting `distinct_source_systems == 0`.

Verified locally:
- `python -m coverage run -m pytest tests/test_ingest.py -q` → 41 passed
- `python -m coverage report --include='philanthropy/ingest/_constituent_events.py' --fail-under=100` → 100% coverage, no missing lines
- `flake8`, `mypy`, and `--doctest-modules` all clean

Added a CHANGELOG entry under [Unreleased] and myself to CONTRIBUTORS.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 ## [Unreleased]
 
 ### Added
+- Test coverage for `constituent_events_to_features`: the all-unparseable-timestamps empty-frame path and the `distinct_source_systems` default-to-zero path when `sourceSystem` is absent from the input. (#51)
 - `AGENTS.md`: every change, including maintainer- and agent-authored ones, must
   go on a branch and through a PR — no direct commits to `main`, no self-merges.
 - `tests/test_no_network.py` enforces in CI what the docs now promise: the package
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -24,7 +24,10 @@ contribution. Code, docs, tests, and review all count.
   [#42](https://github.com/PhilanthroPy-Project/PhilanthroPy/pull/42),
   [#43](https://github.com/PhilanthroPy-Project/PhilanthroPy/pull/43),
   [#44](https://github.com/PhilanthroPy-Project/PhilanthroPy/pull/44)).
-
+- [@AsavariCharati](https://github.com/AsavariCharati) — added test coverage for
+  two untested guards in `constituent_events_to_features`: the all-unparseable-
+  timestamps empty-frame path and the `distinct_source_systems` default when
+  `sourceSystem` is absent.
 ## Getting listed
 
 Add yourself here in the same pull request as your change — one line, your name
--- a/tests/test_ingest.py
+++ b/tests/test_ingest.py
@@ -103,6 +103,15 @@ def test_distinct_source_systems():
     feats = constituent_events_to_features(events)
     assert feats.loc["a@x.edu", "distinct_source_systems"] == 2
 
+def test_distinct_source_systems_defaults_to_zero_when_column_absent():
+    events = [
+        {"constituentEmail": "a@x.edu",
+         "createdAt": "2025-01-01T00:00:00Z", "amount": 10.0},  # no sourceSystem
+        {"constituentEmail": "a@x.edu",
+         "createdAt": "2025-02-01T00:00:00Z", "amount": 20.0},  # no sourceSystem
+    ]
+    feats = constituent_events_to_features(events)
+    assert feats.loc["a@x.edu", "distinct_source_systems"] == 0
 
 # --------------------------------------------------------------------------- #
 # Identity resolution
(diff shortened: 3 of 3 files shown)


shivamlalakiya (approved):
Thanks for filling in these guard-branch tests, @AsavariCharati, clean and matches the existing helper patterns. Appreciate the coverage/lint verification notes in the description too.

shivamlalakiya:
@AsavariCharati thanks again for #74. Those guard branches in `constituent_events_to_features` sat untested for months before you covered them, and guard branches are precisely the code that only ever runs when someone is already having a bad day.

AsavariCharati:
Thanks @shivamlalakiya so a few things tripped me up, all Windows-specific:

- `make ci` / `make riskcov` don't work out of the box since `make` isn't available on Windows by default — I ended up running each command from the Makefile manually (flake8, mypy, doctest, pytest, coverage) one at a time.
- `sh scripts/install_hooks.sh` also doesn't run in PowerShell so  had to use Git Bash for that one step.
- The trickiest one: `pytest --cov=...` (via pytest-cov) threw `ImportError: cannot load module more than once per process` on numpy import, every time. Plain `pytest` and `coverage run -m pytest` both worked fine though  , so I just used `python -m coverage run -m pytest tests/test_ingest.py -q` followed by `coverage report --include=...` instead, and that gave the same result.

Nothing blocking, just extra steps I had to work out. I'll check #188 and see if any of this is already covered  , happy to open something concrete if not!

shivamlalakiya:
Answered at length on #188 rather than twice: https://github.com/PhilanthroPy-Project/PhilanthroPy/discussions/188#discussioncomment-18351685

Short version. The `make` and `sh scripts/install_hooks.sh` friction is filed as #195, labelled `good first issue`, crediting you; the fix is documentation, not a `make.bat`. The `pytest --cov` / numpy `ImportError: cannot load module more than once per process` one I have asked you to file, because I cannot reproduce it and the diagnostic detail is all on your machine. #188 lists exactly what to include.

Thanks for coming back to a closed issue to answer a question. That is not the default.

PR #72, opened 2026-08-13
fix: collapse duplicate JOSS paper, point draft-pdf.yml at the current one

shivamlalakiya:
## Summary
- Two `paper.md`/`paper.bib` pairs were tracked at once: the repo-root pair (current, last touched 2026-08-11 — different author affiliation and a newer, DOI-bearing bibliography) and a stale copy in `paper/` (2026-08-01).
- `.github/workflows/draft-pdf.yml` filters on `paper/**` and points `paper-path` at `paper/paper.md`, so it has only ever built the **stale** draft. The current one has never produced a PDF and never would have until this changed.

## Why
Found during a 2026-08-13 audit, ahead of a JOSS submission. Verified before fixing:
```
$ git log -1 --format="%ai" -- paper.md         # root — canonical
2026-08-11 17:20:16 -0500
$ git log -1 --format="%ai" -- paper/paper.md   # stale
2026-08-01 17:01:43 -0500
```
Confirmed every `@key` citation `paper.md` uses (`lalakiya2025`, `numpy`, `pandas`, `scikit-learn`) exists in the root `paper.bib` before deleting the other copy. Grepped the repo for every other reference to `paper/` or the paper files — only `draft-pdf.yml` referenced the stale path.

## Changes
- Deleted `paper/paper.md` and `paper/paper.bib`.
- `.github/workflows/draft-pdf.yml`: trigger paths and `paper-path` now point at the root `paper.md`/`paper.bib`; the uploaded artifact path updated from `paper/paper.pdf` to `paper.pdf` (the action writes the PDF alongside the source `.md` by default).
- `CHANGELOG.md`: entry under `[Unreleased] / Fixed`.

## Test plan
- [x] `mkdocs build --strict` — exit 0, no new warnings
- [x] `pytest tests/ --collect-only -q` — 1690 collected, no change (no test referenced `paper/`)
- [x] Grepped `tests/`, `mkdocs.yml`, `MANIFEST.in`, `RELEASING.md`, all other workflow files for `paper` — no other stale reference found
- Not independently runnable here: the `draft-pdf.yml` job itself only triggers on paths matching `paper.md`/`paper.bib` on push/PR — will validate on this PR's own CI run

PR #77, opened 2026-08-20
docs: correct four docstrings that described behaviour the code lacks

shivamlalakiya:
No behaviour changes. Four docstrings documented an intent the code never implemented, which is worse than an undocumented gap: a reader who trusts the docstring writes a bug.

| Where | Said | Does |
|---|---|---|
| `FiscalYearTransformer` (class + `transform`) | *appends* `fiscal_year`/`fiscal_quarter` | returns **only** those two columns, drops the input |
| `EncounterTransformer.fit` | "prevents temporal data leakage" | only guarantees nothing from `X` enters the summary |
| `WealthScreeningImputerKNN.group_col_idx` | per-group stratified imputation "improving local accuracy" | stored, never read |
| `GratefulPatientFeaturizer` weights | "commonly-cited AMC development benchmarks" | no published source exists |

## Detail

**`FiscalYearTransformer` is the one with teeth.** `transform` returns a `(n_samples, 2)` block and discards every input column. `get_feature_names_out` has always reported exactly `["fiscal_year", "fiscal_quarter"]`, so the docstring was the outlier. Anyone chaining this into a `Pipeline` on the strength of "appended" silently lost their features with no error. Documented the real contract and pointed at `ColumnTransformer(remainder="passthrough")` / `FeatureUnion` for keeping the originals. Changing the return shape instead would be a Tier-1 break, so the docs moved.

**`EncounterTransformer.fit`** genuinely does freeze the summary independently of `X`, so `transform` is idempotent. What it does *not* do is bound the encounter table by an as-of date: a gift dated 2020 is scored against encounters recorded in 2024 if the table contains them, and `days_since_last_discharge` is computed from the all-time max discharge. The claim is now scoped to what holds, with a warning naming what does not. Actually fixing the cutoff is a behaviour change and belongs in its own PR.

**`group_col_idx`** is documented, accepted, stored, and never read. Left in the signature so existing calls do not start raising, but the docstring now says "ignored" and a test asserts passing it changes nothing. Whether to implement or remove it is an API decision, so I have filed it as an issue rather than picking one here.

**The service-line weights** (cardiac 3.2, oncology 2.9, neuroscience 2.7) were attributed to "commonly-cited AMC development benchmarks". No such source is cited anywhere and I could not find one. They are placeholders encoding an ordering, and the comment now says that plainly and says not to cite them. `docs/reference/index.md` already described them as "illustrative, not benchmarked", so this brings the code comment in line with the docs rather than the other way round.

Also dropped the claim that KNN imputation "materially outperforms" median imputation. No benchmark in this repository measures it.

## Tests

`tests/test_documented_contracts.py` (new file, 3 tests) locks each corrected contract, so the next rewording has to match a passing test rather than an intention. New file rather than an append to an existing one, deliberately: five PRs are in flight and this avoids conflicting with any of them.

## Verification

    make ci        # exit 0, 1693 passed, 2 skipped
    make riskcov   # 95% against the 93% floor

Overall coverage 95.66% against the 92% floor.


shivamlalakiya:
Heads-up on overlap: #90 covers the same four docstring corrections independently, bundled with a 431-em-dash sweep across 91 files. #90 is currently `CONFLICTING` (98 files) because it also re-applies the conformal work merged in #75 and the paper collapse merged in #72.

This PR is `MERGEABLE` and scoped to the docstrings only. It also adds `tests/test_documented_contracts.py`, which #90 does not have: three tests that lock the corrected contracts, so the next rewording has to match a passing test rather than an intention. That matters most for `FiscalYearTransformer`, where the docstring and `get_feature_names_out` disagreed for a while and only the docstring was wrong.

Two corrections in #90 that this PR does **not** have and that should survive either way: `days_since_last_discharge` is documented as an integer but is `float64`, and `encounter_frequency_score` is documented as a count of distinct records but is `log1p` of the row count.

No preference on which becomes the vehicle. Flagging so the two do not get merged into each other.

PR #76, opened 2026-08-20
docs: collapse the duplicate JOSS paper and settle its metadata

shivamlalakiya:
Stacked on #75 so the `paper.md` hunks do not collide. **Base is `feat/conformal-pvalue`; GitHub retargets this to `main` automatically once #75 merges.** Review #75 first.

> **Supersedes #72.** That PR does the paper collapse and the `draft-pdf.yml` repoint, both of which are included here. Two open PRs editing the same four workflow lines and deleting the same two files will conflict, so please close #72 in favour of this one (or close this one and I will rebase the rest onto #72, whichever you prefer).

## What this fixes

**Two paper drafts were tracked at once.** `paper.md`/`paper.bib` at the repo root, and a divergent 2026-08-01 copy in `paper/` with a different title and a different affiliation. `draft-pdf.yml` filtered on `paper/**` and set `paper-path: paper/paper.md`, so it built only the stale copy and the draft that is actually maintained had never produced a PDF. Deleted `paper/`, and repointed the workflow including its `paths` filter, so edits to the root draft now trigger a build.

**Two citations were missing.** The Statement of need asserts temporal leakage is the endemic failure mode of donor data with nothing behind it, and the AMC domain claim had no ethics reference. Restored `kaufman2012leakage`, `kapoor2023leakage` and `collins2018grateful`, which the `paper/` copy carried and the root draft had dropped.

**The affiliation disagreed three ways.** `paper.md` said Independent Researcher, `paper/paper.md` said Northeastern University, `.zenodo.json` said Washington University in St. Louis. Settled on "Independent Researcher" per the maintainer. This matters beyond tidiness: `.zenodo.json` is minted into a permanent citable record, and a JOSS reviewer finds a three-way affiliation conflict in about five minutes.

## Verification

    make ci        # exit 0, 1700 passed, 2 skipped
    make riskcov   # 95% against the 93% floor

Overall coverage 95.70% against the 92% floor. Every `@key` cited in `paper.md` resolves to an entry in `paper.bib` and there are no unused entries, checked both directions.

## Not in this PR

The version metadata still disagrees three ways (`pyproject.toml` and `CITATION.cff` say 1.0.0, PyPI latest is 0.6.0, newest tag is `v0.6.0`, CHANGELOG has `## [1.0.0] - TBD`). That is a release decision, not a docs fix, so nothing here touches a version number. The stale `CITATION.cff` DOI comment is blocked on the same decision.

No CONTRIBUTORS.md entry needed (maintainer already listed).

Issue #85, opened 2026-08-20
group_col_idx: implement per-group KNN imputation or deprecate the parameter


shivamlalakiya:
## Current state

`WealthScreeningImputerKNN.group_col_idx` is documented, accepted, stored, and **never read**. The only three occurrences in the package are the docstring, the `__init__` signature, and the assignment:

- `philanthropy/preprocessing/_share_of_wallet.py` — docstring
- same file — `group_col_idx: Optional[int] = None` in `__init__`
- same file — `self.group_col_idx = group_col_idx`

The docstring previously promised behaviour with a performance claim attached:

> Column index of a group variable (e.g., zip-code encoded as int) to stratify KNN imputation. When provided (and `strategy="knn"`), imputation is performed independently per group, improving local accuracy.

PR #77 corrected the docstring to say it is ignored and added a test asserting that passing it does not change the output. This issue is the underlying decision, which is an API call rather than a docs fix.

## The decision

**A: implement it.** Group the rows by the column at `group_col_idx` and run `KNNImputer` independently within each group, falling back to the global fit for groups too small to have `n_neighbors` complete rows. This is the behaviour already promised and it is plausibly useful: wealth-screening appends genuinely do cluster geographically.

**B: remove it.** `philanthropy.preprocessing` is Tier 1 per `docs/reference/index.md`, so removal needs a `DeprecationWarning` runway from 1.0.0. Nobody can depend on behaviour that never existed, but callers may be *passing* the argument, and having it start raising is a break.

Related, and part of the same question: at default settings `WealthScreeningImputerKNN` is equivalent to a bare `sklearn.impute.KNNImputer` plus the missingness indicator. `group_col_idx` is the main thing that would give the class a reason to exist beyond that wrapper, which argues for A.

## If A

- Guard the small-group case explicitly; `KNNImputer` with fewer complete rows than `n_neighbors` silently degrades.
- Keep it leakage-safe: group statistics must be frozen at `fit`, and a group unseen at `fit` time must fall back to the global imputer rather than being fitted at `transform` time. That is the non-negotiable contract in `AGENTS.md`.
- Ignore it (with the existing documented behaviour) when `strategy != "knn"`.

PR #90, opened 2026-08-20
Remove AI-tell artifacts: em dashes, wrong docstrings, unsourced constants

shivamlalakiya:
Three things in this repo read as machine-written to a reviewer skimming for it. Two of them are also just wrong. This fixes all three, plus the JOSS paper prep and `conformal_pvalue` that were already sitting uncommitted on this branch.

## 1. Em dashes

431 of them across 91 files: `paper.md`, `README.md`, all 43 docs pages, docstrings, inline comments, `CHANGELOG.md`, the `Makefile`, `.flake8` and both CI workflows.

Each site got the punctuation the sentence actually wanted, chosen individually rather than by blanket substitution:

- label or definition -> colon (`## Local gate: exact commands`)
- appositive -> commas (`Calibrated HistGradientBoostingClassifier, NaN-native`)
- two independent clauses -> semicolon (`reject NaN; impute internally with frozen per-column medians`)
- paired dashes around a list -> parentheses (`All fitted statistics (fill values, summaries, coefficients) are computed from TRAINING data`)

The only em dash left in the repository is inside a nonprofit's name in `philanthropy/datasets/data/ciob_official_fundraising.csv`. That is source data, not prose.

## 2. Docstrings describing code that does not exist

This is the failure mode worth taking seriously: documentation written from intent, where nobody followed the code afterwards.

**`FiscalYearTransformer`** said `transform` appends `fiscal_year` and `fiscal_quarter` to `X`. It returns only those two columns and drops everything else (`_transformers.py`, `out_df` is built fresh). The class and method docstrings now say that, document the `NaN` paths, and point at `ColumnTransformer` for the append case.

**`EncounterTransformer.fit`** claimed that deriving `encounter_summary_` from `encounter_df` alone "prevents temporal data leakage when the transformer is placed **before** a time-based train/test split inside a pipeline." It does not. What it actually guarantees is that nothing from `X` or `y` enters the summary, so refitting inside a CV fold cannot leak gift-level information across folds. But the summary takes `max(discharge_date)` over every row of `encounter_df` regardless of date, so encounters recorded after an evaluation cut-off reach both the training and the test rows at transform time. The caller has to filter `encounter_df` to the training window. The docstring now states the real guarantee and the real gap, in a library whose headline claim is leakage safety.

Same class, two smaller ones: `days_since_last_discharge` was documented as an integer (it is `float64`, so it can carry `NaN`), and `encounter_frequency_score` as a count of *distinct* encounter records (it is `log1p` of the row count).

## 3. Plausible constants with plausible provenance and no source

The comment above `_SERVICE_LINE_CAPACITY_WEIGHTS` credited the ordering to "commonly-cited AMC development benchmarks" and cited none. That is the exact shape of a confabulated citation, whether or not the numbers were chosen by hand.

There is no source. They are hand-picked placeholders so `use_capacity_weights=True` has something to run with out of the box. The comment now says so, tells the reader to treat the values as a shape rather than as values, and points at the how-to page. The `capacity_weights` and `use_capacity_weights` parameter docs were updated to match.

## Also: a dead parameter

`WealthScreeningImputerKNN(group_col_idx=...)` was stored in `__init__`, read nowhere, and documented as performing per-group KNN imputation "improving local accuracy" that was never implemented. Removed. Nothing in the library, tests, docs or README passed it, so no behaviour changes; passing it now raises `TypeError` instead of being silently ignored.

## Not addressed

The fourth artifact, volume against timeline, cannot be fixed by editing files. `git log` is 50 commits in July and August 2026 (there is no April-to-June gap; history simply starts in July), against 43 docs pages, ~8k lines of tests, and a `CHANGELOG.md` covering 0.1.0 through 1.0.0. Rewriting history to spread those dates would be falsification, not cleanup. The honest options are to disclose AI assistance in the paper or repo, or to leave it and let the work stand. That is a call for the maintainer, not this PR.

## Verification

```
make ci        1700 passed, 2 skipped, coverage 95.70% (floor 92%), exit 0
make riskcov   exit 0
```

Changelog entries added under `## [Unreleased]` in **Changed** and **Removed**.

## First commit on this branch

`6ddc8ca` is the paper-prep and `conformal_pvalue` work that was already uncommitted in the working tree; it is included here rather than stranded. It adds `philanthropy.metrics.conformal_pvalue`, deletes the stale duplicate `paper/` draft, restores the leakage and grateful-patient-ethics citations, and settles the affiliation across `paper.md` and `.zenodo.json`.


shivamlalakiya:
Flagging an overlap before anyone merges this, since two sessions worked the same uncommitted branch in parallel and split it differently.

**This PR is now `CONFLICTING` (98 files) because two thirds of it already landed.**

- The `conformal_pvalue` work merged as #75.
- The duplicate-paper collapse and the `draft-pdf.yml` repoint merged as #72.

So those sections of this PR are re-applying commits that are already on `main`, which is where the conflict comes from.

**The docstring section duplicates #77**, which is `MERGEABLE` right now. #77 covers the same four items independently: `FiscalYearTransformer` append-vs-replace, the `EncounterTransformer.fit` leakage claim, `group_col_idx` being documented and never read, and the service-line weights attributed to "commonly-cited AMC development benchmarks" that have no source. It also adds `tests/test_documented_contracts.py` so each corrected contract is locked by a passing test rather than by a reviewer's memory.

Two details in your section 2 that #77 does not have and that are worth keeping: `days_since_last_discharge` documented as an integer when it is `float64`, and `encounter_frequency_score` documented as a count of *distinct* records when it is `log1p` of the row count. Those are real and should survive whichever way this goes.

**What is still genuinely missing from `main`, and what #76 covers:**

```
$ git show origin/main:.zenodo.json | grep affiliation
      "affiliation": "Washington University in St. Louis"
$ git show origin/main:paper.bib | grep -oE '^@[a-z]+\{[^,]+' | sed 's/.*{//'
scikit-learn pandas numpy lalakiya2025 vovk2005conformal
```

The affiliation still disagrees with `paper.md`, and `kaufman2012leakage` / `kapoor2023leakage` / `collins2018grateful` are still absent, so the temporal-leakage claim in the Statement of need and the AMC domain claim both still have nothing behind them. #76 is `MERGEABLE` and fixes exactly those two things.

**Suggested resolution**, entirely your call:

1. Merge #76 (affiliation + citations) and #77 (docstrings + contract tests) as they are, both mergeable today.
2. Rebase this PR onto `main` and reduce it to its unique contribution: the 431-em-dash sweep across 91 files, plus the two `EncounterTransformer` type/semantics corrections noted above.

That leaves one PR per concern and no PR re-applying merged work. Happy to do the reduction on this branch myself instead if you would rather keep this one as the vehicle and close #76/#77 — say which and I will do it rather than both of us editing the same files again.

shivamlalakiya:
Superseded by #91, which is this branch reduced to the em-dash sweep as the maintainer asked.

#91 carries the 442 dashes across 91 files and **takes the replacement choices from this branch rather than re-deriving them**, so the per-site judgement here is preserved rather than redone with a blanket substitution. Nothing else from here is in it: the docstring corrections landed in #77, the paper collapse in #72, `conformal_pvalue` in #75.

Two things I changed while extracting it, both noted in #91:

- The `paper.md` hunk here also added the `[@collins2018grateful]` citation, which belongs to #76. #91 drops that and asserts the `@key` set in `paper.md` is byte-identical to `main`.
- I kept your `tests/test_preprocessing.py` change from `— mathematically guaranteeing` to `to check`, even though it is more than punctuation. Property-based tests do not mathematically guarantee anything, so that was a real fix worth keeping.

Also worth carrying over separately, since #91 is punctuation-only and #77 did not cover them: the two `EncounterTransformer` corrections from your section 2, that `days_since_last_discharge` is documented as an integer but is `float64`, and that `encounter_frequency_score` is documented as a count of distinct records but is `log1p` of the row count. Both are still wrong on `main`. Happy to send a two-line PR for those if you want them.

Recommend closing this once #91 lands.

shivamlalakiya:
Closing per review, superseded rather than merged.

The review confirmed three blockers that make merging this actively regressive against current `main`, not merely redundant:

1. It deletes `paper/paper.md` but never repoints `.github/workflows/draft-pdf.yml`, which still sets `paper-path: paper/paper.md`. Its one CI check fails for exactly that reason: `realpath: paper/paper.md: No such file or directory`.
2. Removing the `WealthScreeningImputerKNN(group_col_idx=...)` parameter breaks `tests/test_documented_contracts.py::test_group_col_idx_is_inert`, which #77 added to `main` after this branch was cut. The "1700 passed" in the description was measured pre-#77.
3. That removal is an undeprecated breaking change to a Tier 1 stable symbol, filed under `### Removed` with self-contradictory text ("changes no behaviour ... now raises `TypeError`") and no major bump.

It also reverts two docstring corrections already merged via #71 and #92.

**Nothing is lost.** Every unique contribution here is carried elsewhere, and I deliberately waited until those landed before closing:

- The em-dash sweep, rebased and green: #91 (it takes this branch's per-site replacement choices rather than re-deriving them, so the judgement here is preserved).
- The `_encounters.py` docstring corrections: #97.
- The three `paper.bib` entries and their `paper.md` citations: #76.
- The `.zenodo.json` affiliation change: #76, byte-identical.
- The `days_since_last_discharge` / `encounter_frequency_score` corrections this branch spotted: #97, with tests.

The `group_col_idx` removal is the one thing not carried, and deliberately: `main` documents it as a no-op locked by #77's test, and #99 implements it instead. Worth noting that measuring it since made the weaker case, so deprecating it later per issue #85 option B is still reasonable.

Thanks for the sweep itself, which found real defects the narrower PRs then fixed properly.

Issue #63, opened 2026-08-11
test: `EncounterRecencyTransformer` coverage gap — timezone/overflow/ndarray-fallback branches untested (82% coverage, worst in risk tier)


shivamlalakiya:
## Summary

`philanthropy/preprocessing/_encounter_recency.py` is at **82% branch
coverage**, the lowest of any file in the risk tier `make riskcov` enforces at
a 93% floor — this file alone is dragging that subtree's average down. Unlike
the other coverage-gap issues in this repo, which name one or two specific
guards, this one is a cluster spanning both `fit` and `transform` on
`EncounterRecencyTransformer`, plus its private helpers. Filing it as one
issue rather than several because the branches share a theme (timezone- and
NaN-handling edge cases) and a contributor working through them will end up
reading the whole `fit`/`transform` pair anyway.

Untested regions, grouped by what triggers them (confirm current line numbers
with the coverage command below — they will drift as other work touches this
file):

1. **`_parse_dates`'s tz-localise fallback.** `parsed.dt.tz_convert(self.timezone)`
   is wrapped in `try/except Exception`, falling back to
   `parsed.dt.tz_localize(self.timezone)`. No test parses a tz-naive series
   through a `timezone`-configured transformer, so the fallback never runs.
2. **`_compute_recency_features`'s tz-mismatch branches.** The three-way branch
   comparing `dates.dt.tz` against `ref.tzinfo` (tz-aware dates vs tz-naive
   reference, and the reverse) — only the "both match" case is exercised today.
3. **`fit`'s reference-date inference paths.** Both `UserWarning` branches —
   "no parseable dates found in training data" and "`X` is not a DataFrame" —
   defaulting `reference_date_` to today.
4. **`transform`'s ndarray-fallback and missing-date-column paths.** The
   branch that rebuilds a DataFrame from `feature_names_in_` when `X` isn't
   already one, the "cannot resolve column names" branch returning an
   all-`NaN` array, and the per-column "date column not found in X" warning
   inside the `for col in cols` loop.
5. **`_days_since_day_resolution`**, the overflow-safe fallback for date spans
   beyond ~292 years — entirely unreached, since no test constructs dates that
   overflow `int64` nanosecond resolution.

### Where

| File | Locate with |
|---|---|
[...]

PR #106, opened 2026-08-21
docs: add NumPy Parameters/Returns sections to the _scoring KPI functions

### LINKED ISSUE #60: docs: four KPI functions in `philanthropy.metrics._scoring` document neither Parameters nor Returns
## Summary

All four functions in `philanthropy/metrics/_scoring.py` — `donor_retention_rate`,
`donor_acquisition_cost`, `cost_per_dollar_raised`, `fundraising_roi` — have a
prose summary and a short explanatory paragraph, but none has a formal
`Parameters` or `Returns` section. The API reference is generated by
mkdocstrings from a bare `::: philanthropy.metrics` directive, so each of
these four renders with its signature and the prose, but the arguments and
return value are not itemised the way every other metric in the package
documents them (see `gift_concentration_gini` or `top_donor_share` in
`philanthropy/metrics/_concentration.py` for the shape to copy).

This is a good first issue: no logic changes, and the prose already explains
the behaviour — the work is reformatting it into NumPy sections.

### Where

| File | Locate with |
|---|---|
| `philanthropy/metrics/_scoring.py` | `grep -n "^def " philanthropy/metrics/_scoring.py` |

### What to change

NumPy format (`mkdocs.yml` sets `docstring_style: numpy`). For each of the four
functions, add:

- `Parameters` — every keyword-only argument, one line each.
- `Returns` — the value and, where the function has one, the `np.inf` sentinel
  case already described in the existing prose (e.g. "`np.inf` when
  `new_donors_acquired` is 0").

Keep the existing prose as the summary line(s) above the new sections; don't
delete the explanatory context, just add the missing structure around it.

### Tests to add or extend

None — doc-only change, no new branch to cover.

### Done when

```bash
python -c "
import inspect, sys, philanthropy.metrics as M
missing = [
    n for n in ('donor_retention_rate', 'donor_acquisition_cost',
                'cost_per_dollar_raised', 'fundraising_roi')
    if 'Parameters' not in (getattr(M, n).__doc__ or '')
    or 'Returns' not in (getattr(M, n).__doc__ or '')
]
print('MISSING:', missing or 'none'); sys.exit(1 if missing else 0)
" && echo OK
```

slegarraga:
Closes #60.

Adds formal `Parameters` and `Returns` sections (NumPy style, matching `_concentration.py`) to `donor_retention_rate`, `donor_acquisition_cost`, `cost_per_dollar_raised`, and `fundraising_roi` so mkdocstrings itemises them like every other metric. The existing prose summaries are unchanged; no logic touched. flake8 clean.

### DIFF
--- a/philanthropy/metrics/_scoring.py
+++ b/philanthropy/metrics/_scoring.py
@@ -15,6 +15,20 @@ def donor_retention_rate(
 
     Returns a fraction in ``[0.0, 1.0]``; ``0.0`` when ``prior_donors`` is
     empty (no base to retain from).
+
+    Parameters
+    ----------
+    current_donors : collection
+        Donor identifiers seen in the current period.
+
+    prior_donors : collection
+        Donor identifiers seen in the prior period.
+
+    Returns
+    -------
+    float
+        Retained share in ``[0.0, 1.0]``. Returns ``0.0`` when
+        ``prior_donors`` is empty (no base to retain from).
     """
     current_set = set(current_donors)
     prior_set = set(prior_donors)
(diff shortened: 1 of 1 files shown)


slegarraga:
Merged within the hour — thank you for the quick turnaround! One more incoming: #107 covers the encounter-recency test gap (#63) with the same measure-first approach.

PR #123, opened 2026-08-21
fix: reject network-scheme paths at the read boundary

### LINKED ISSUE #114: "No network calls" claim is broader than the code: encounter_path and CLI data path accept remote URIs
### What's wrong

`README.md:142` and `docs/explanation/security_review_answers.md:16` both
state PhilanthroPy "makes no network calls of any kind." `tests/test_no_network.py`
enforces this by poisoning every socket, but it only exercises
`generate_synthetic_donor_data` plus a local CRM ingest, so it never proves the
claim against the public parameters that actually accept a path.

Three call sites accept a bare path string and hand it to a pandas reader that
supports remote URIs (`https://`, `s3://`, `gs://`) out of the box:

| File | Locate with |
|---|---|
| `philanthropy/cli.py` | `grep -n "pd.read_csv(path)" philanthropy/cli.py` |
| `philanthropy/preprocessing/_grateful_patient.py` | `grep -n "pd.read_parquet(self.encounter_path)" philanthropy/preprocessing/_grateful_patient.py` |
| `philanthropy/preprocessing/_encounters.py` | `grep -n "pd.read_parquet(self.encounter_path)" philanthropy/preprocessing/_encounters.py` |

`encounter_path` and the CLI's `args.data` are documented public parameters,
not internal plumbing. This is exactly the security-review page's stated
audience: an institutional reviewer who reads "no network calls of any kind"
and then finds a parameter that accepts `s3://...`.

### What to change (pick one, it's a decision not a bug fix)

1. **Reject non-local paths at the boundary.** In `CRMCleaner`/`cli.py` and the
   two encounter transformers, validate that the path has no network scheme
   (`urllib.parse.urlparse(path).scheme not in ("", "file")`) and raise
   `ValueError` before calling the pandas reader. Keeps the claim true as
   written.
2. **Re-scope the wording.** Change "no network calls" to "no network calls of
   its own" (or similar) in both `README.md` and
   `docs/explanation/security_review_answers.md`, and add one sentence noting
   that user-supplied paths are passed through to pandas as-is and can be
   `https://`/`s3://`/`gs://` if the caller chooses.

Option 1 is stronger and matches what the docs already claim; option 2 is a
smaller diff. Either is fine, but the current state (docs claim option 1,
code does neither) is the one combination that isn't.

### Tests to add or extend

- File: `tests/test_no_network.py`
- If option 1: add `test_encounter_path_rejects_remote_scheme` and
  `test_cli_data_path_rejects_remote_scheme`.
- If option 2: no test change needed beyond the doc wording; consider a
  regression test that documents the accepted behavior explicitly rather than
  leaving it implicit.

slegarraga:
Closes #114 — implemented **option 1** (keep the claim true as written).

**What**

- New `ensure_local_path()` helper in `philanthropy/utils/_validation.py`: raises `ValueError` when a path carries any scheme outside `''`/`'file'`.
- Wired into all three boundary sites from the issue: `cli._read_csv`, `GratefulPatientFeaturizer.fit`'s `encounter_path` read, `EncounterTransformer.fit`'s `encounter_path` read.
- `tests/test_no_network.py` gains the two tests the issue names (`test_encounter_path_rejects_remote_scheme` covering both transformers × https/s3, `test_cli_data_path_rejects_remote_scheme` covering https/gs) plus a control test proving local paths still load under the poisoned-socket fixture.

**Rationale for option 1 over 2:** the security-review page's audience is institutional reviewers evaluating a HIPAA-safe tool; silently supporting `s3://` contradicts the product's core claim, and this project documents that 0.x minors may break — so a behavioral tightening is acceptable now and would only get harder post-1.0. If you'd prefer option 2's smaller diff I'm happy to swap.

Behavior note: this is technically breaking for anyone intentionally reading from object storage via these parameters; worth a line in the next release notes.

Full suite green locally (1810 passed); flake8 clean.


shivamlalakiya:
Thanks for this @slegarraga, and sorry for editing your branch directly rather than asking. The guard is right and I have not touched it. I pushed one commit (0612dce) that changes the *claim* it cites, and the reasoning is worth writing out because it is not obvious from the diff.

### The guard is correct and stays

`ensure_local_path` is called on exactly three things: `EncounterTransformer(encounter_path=...)`, `GratefulPatientFeaturizer(encounter_path=...)` and the CLI's `--data`. All three are paths a user supplies for their own donor data, and none of them should ever be remote. That is the real gap #114 named, and you closed it properly: the guarantee held for the library's own logic and was false for its documented public parameters.

### What was wrong was the claim, and it was unenforced

The docstring gave the reason as "PhilanthroPy makes no network calls of any kind", and `README.md` said "the package imports no HTTP client at all". Both absolute, and nothing checked them across the package. `tests/test_no_network.py` poisons sockets only inside the four tests that request the `no_network` fixture, so a downloader added to a module no test imports would leave every one of those tests green while three public pages quietly became false.

That is not hypothetical. #124 validates the library against real donor data, which needs a KDD Cup 1998 fetcher in `philanthropy/datasets/`. Nothing in your diff blocks it, but the sentences around it would have started lying the day it landed, and no test would have said so.

### What the commit does

Two precise guarantees instead of one blanket one, and the previously-unenforced half is now machine-checked:

- **"Never transmits your data"** is the claim an institutional review actually cares about, and a read-only public-dataset fetcher does not touch it.
- **"Downloads nothing"** is now enforced by `test_no_module_imports_a_network_client`, which `ast`-parses every module in the package and fails the build if one imports a network-capable library without appearing on `_NETWORK_ALLOWED`.

`_NETWORK_ALLOWED` is **empty**, so the effective promise is unchanged and is now enforced across modules no test happens to import, rather than only on the paths the socket fixture walks. Adding an entry requires updating `README.md`, `SECURITY.md` and the security review Q&A in the same PR. `urllib.parse` is deliberately not on the forbidden list, since it is what your `ensure_local_path` uses to reject remote paths.

I also added question 1a to the security review Q&A documenting your remote-path rejection. That page exists to be forwarded to a privacy officer, and "what if someone passes it an `s3://` path?" is the first follow-up to question 1, so the behaviour belongs there rather than only in the CHANGELOG.

### Verified

`make ci` exit 0 (1855 passed, 2 skipped, coverage 96.74% against the 92% gate), `make riskcov` exit 0, `mkdocs build --strict` exit 0. Both new tests were confirmed to fail when they should: a probe module importing `urllib.request` is caught, and a stale allowlist entry is caught.

If you would rather own this wording yourself, say so and I will drop my commit and take it in a separate PR.

**I am not merging this.** Per AGENTS.md that is a human's call, and there is one more reason to wait a few days: 90 of 171 commits on `main` are in August, against JOSS's desk-reject criterion on development "concentrated into a few days or weeks" (#127). Merging is not free this month.

shivamlalakiya:
`ensure_local_path()` uses `urlparse(path).scheme`, and on Windows that misreads an absolute drive-letter path as a network scheme:

```python
>>> from urllib.parse import urlparse
>>> urlparse(r"C:\Users\x\file.csv").scheme
'c'
>>> urlparse("C:/Users/x/file.csv").scheme
'c'
```

So `encounter_path="C:\data\encounters.parquet"` or `--data C:\gifts.csv` would raise `ValueError` on Windows, even though `pyproject.toml` still classifies this package `Operating System :: OS Independent`. CI doesn't catch it since there's no `windows-latest` job and the new tests only cover POSIX-style local paths plus the remote schemes.

Worth guarding on drive-letter schemes (single ASCII letter) before the reject, or checking `len(scheme) > 1` alongside the existing allowlist. Holding off merging until this is addressed.

PR #128, opened 2026-08-21
ci: block merges that skip CHANGELOG.md or contributor credit

### LINKED ISSUE #113: CI: make CONTRIBUTORS.md/CHANGELOG.md updates a merge-blocking check
### What's wrong

The credit process has broken twice: PR #45 had to backfill `CHANGELOG.md`
and `CONTRIBUTORS.md` for #40/#42/#43/#44 after the fact, and #64/#65/#66
merged 2026-08-11 with neither (see the companion issue crediting
`@shubhrai23`). Nothing currently stops a PR from merging without one or both
updates.

### Where

| File | Locate with |
|---|---|
| `.github/workflows/ci.yml` | `grep -n "^jobs:" .github/workflows/ci.yml` |

### What to change

Add a CI check (new job or step in the existing `ci.yml`) that, on a pull
request, fails if the diff touches source files under `philanthropy/` but
does **not** touch `CHANGELOG.md`. Same check, or a separate one, for
`CONTRIBUTORS.md` when the PR author's GitHub handle does not already appear
in it. Use `git diff --name-only "$(git merge-base origin/main HEAD)" HEAD`
(or the `github.event.pull_request` base/head refs) to compute the changed
file list; skip the check for PRs opened by the maintainer's own account if
that ends up necessary to avoid blocking release/chore commits that touch no
runtime behavior.

### Done when

- A PR that changes a file under `philanthropy/` without touching
  `CHANGELOG.md` fails CI.
- A PR from a new contributor without a `CONTRIBUTORS.md` line fails CI.
- `make ci` still passes on `main` (the check must not fire retroactively on
  history that predates it).

This is process/tooling work, not a design decision, but it touches CI
workflow internals; scope it as a normal (not "good first issue") task.

slegarraga:
Closes #113.

**What**

- `scripts/check_credit.sh`: given base/head SHAs and the PR author login, fails when the diff touches files under `philanthropy/` without also touching `CHANGELOG.md`, or when the author's handle is absent from `CONTRIBUTORS.md`. Emits GitHub `::error::` annotations so the failure surfaces inline on the Files tab.
- A standalone `credit-guard` job in `ci.yml` wires it up on `pull_request` events only — direct pushes to main and history that predates the guard are never re-litigated, per the issue's done-when.

**Verified locally against real diffs:**

| Scenario | Result |
|---|---|
| Source change, no changelog | ✗ exit 1 |
| Source change, author not in CONTRIBUTORS.md | ✗ exit 1 |
| Source change, changelog + credited author | ✓ exit 0 |
| Tests/docs-only diff | skipped |

Design choices worth flagging:

- The author check keys on the `@handle` string in CONTRIBUTORS.md, matching the file's existing format.
- No maintainer exemption was added: the issue allows one 'if necessary', but since release/chore commits go straight to main they never hit a PR check anyway. Easy to add later if it chafes.
- My own open #123 needed a CHANGELOG entry to comply — updated in that branch already. The guard works on its author first.


shivamlalakiya:
CI is green and the script/workflow wiring looks solid, but per AGENTS.md every PR needs an entry under `## [Unreleased]` in CHANGELOG.md (you're already in CONTRIBUTORS.md, so that part's covered). A bit ironic for the credit-guard PR itself to skip it. Could you add a line describing the new `credit-guard` job? Holding the merge until then.

slegarraga:
Done in c4c159a: added the entry under `## [Unreleased]` describing the
`credit-guard` job (what it gates, how it's wired into `ci.yml`, and that it
runs on `pull_request` events only). Good catch on the irony; the guard now
has nothing left to catch on this one.

shivamlalakiya:
Reviewed. The script and the intent look right, but this cannot merge as-is: the `credit-guard` job has no event guard.

`ci.yml` triggers on `push: branches: [main]` as well as `pull_request`. On a push event `github.event.pull_request` is null, so `BASE_SHA`, `HEAD_SHA` and `PR_AUTHOR` all expand to empty strings, and `scripts/check_credit.sh "" "" ""` runs `git diff --name-only "" ""`, which fails under `set -e`. Every post-merge run on `main` would go red. The CHANGELOG entry already claims the job is "wired into ci.yml on `pull_request` events only" -- that needs to be true in the YAML:

```yaml
  credit-guard:
    name: Changelog and contributor credit (issue #113)
    if: github.event_name == 'pull_request'
    runs-on: ubuntu-latest
```

Two smaller notes, neither blocking:

- `grep -qE "@${author}(...)"` interpolates the login into a regex. Human logins are `[A-Za-z0-9-]` so they are safe, but a bot author (`dependabot[bot]`) would be read as a character class. Bot PRs do not touch `philanthropy/` today, so this is latent, not live.
- The changelog check only asserts that `CHANGELOG.md` was touched, not that the new line landed under `[Unreleased]`. Fine as a guard, worth knowing it is not the stronger check.

Also: GitHub still reports this branch as CONFLICTING, but that state is stale -- `git merge-tree --write-tree origin/main` on the head commit merges clean. Run `gh pr update-branch` (or rebase) after adding the `if:` and it should go green.

slegarraga:
Fixed both points and pushed as 0af3a45 (branch also merged current `main`; GitHub now reports MERGEABLE):

- `credit-guard` carries `if: github.event_name == 'pull_request'`, exactly as sketched, so push builds to `main` skip the job instead of invoking `check_credit.sh` with empty SHAs. The CHANGELOG's "on `pull_request` events only" claim is now literally true in the YAML.
- Closed the latent bot-login note too while there: `check_credit.sh` matches the `@login` token literally (awk `index()` plus a boundary check) rather than interpolating the author into an ERE, so a future `dependabot[bot]` source PR cannot be misparsed as a character class. Kept the error text identical so annotations stay stable.

Verification before pushing:

- `bash -n` clean; workflow parses and asserts equal to your suggested expression.
- A 7-case harness against a scratch repo, all passing: human listed -> OK; prefix trap (`@ana` vs listed `@anakin`) -> correctly rejected; `@dependabot[bot]` listed -> OK via literal match; bot missing -> rejected; diff without `philanthropy/` files -> skipped; missing CHANGELOG -> rejected; empty author -> rejected with the "could not be determined" annotation.
- On CI at this head: Changelog and contributor credit pass (5s), lint/typecheck, CodeQL, build + dist checks and the full test matrix are green.

On your last note, agreed the touched-file check is the weaker form; leaving it as-is for this guard per your framing. And thanks for the sharp catch, every post-merge run on `main` going red would have been an embarrassing debut for issue #113's job. Congrats on landing #140 and #141 as well.

PR #130, opened 2026-08-21
docs: add a reticulate (R) quickstart to the README

### LINKED ISSUE #117: README: add a reticulate (R) quickstart block
### What's wrong

R is the incumbent language among this project's actual audience (Apra's intro
data-science course, CASE's "Introduction to Advancement Analytics Using R" at
DRIVE 2025, and both prior attempts at this exact library,
`michaelpawlus/fundraising_analytics` and `crazybilly/fundRaising`, are R).
README.md has no `reticulate` example showing an R user how to call
PhilanthroPy from R.

### What to add

A short (~10-line) block in `README.md`, after the Python quickstart, showing
a minimal `reticulate` call into one estimator (e.g. `DonorPropensityModel` or
`generate_synthetic_donor_data`). Something in the shape of:

```r
library(reticulate)
philanthropy <- import("philanthropy")
datasets <- import("philanthropy.datasets")

df <- datasets$generate_synthetic_donor_data(n_samples = 1000L, random_state = 0L)
```

### The one thing that will silently break this

`reticulate` sends bare R numerics as doubles. scikit-learn's integer-typed
constructor params (`n_samples`, `random_state`, `n_estimators`, etc.) reject
floats. Every integer argument needs the `L` suffix (`1000L`, not `1000`), or
the call raises inside sklearn's `validate_data`/parameter validation with a
confusing error for someone who doesn't know why.

### Why this needs an R environment to close

**This has not been executed in R by anyone yet.** Whoever picks this up needs
a working R + `reticulate` install to actually run the block before it's
merged; a plausible-looking snippet that's never been executed is worse than
none; if the `L` suffixes are missing or wrong the block fails on the first
person who copies it. Confirm it against whatever `philanthropy` version is
current on PyPI when you pick this up.

### Done when

The `reticulate` block in `README.md` has actually been run against an
installed `philanthropy` and produces output, not just parses.

slegarraga:
Closes #117.

Adds the ~10-line R block after the Python Quick Start: synthetic data → fit → `predict_affinity_score`, plus the `L`-suffix warning the issue flagged.

**Executed, not plausible-looking.** Per the issue's hard requirement, the block was run for real before this PR:

- R 4.6.1 + reticulate against a venv built from this tree
- without `L`: `TypeError: SeedSequence expects int or sequence of ints for entropy not 0.0` — exactly the failure mode described
- with `L`: generator returns 1000×5, model fits, scores span 0–100

The block in the PR is character-for-character what was executed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,30 @@ the threshold from your team's capacity, not from this table.
   <em>Output of <code>plot_affinity_distribution()</code>: the 0-100 affinity scores separate the two groups on average, with tails that overlap. Ranking works; a single clean cut point does not exist.</em>
 </p>
 
+### Using PhilanthroPy from R
+
+Advancement analytics has been an R-first field for years, so PhilanthroPy
+also works from R through [`reticulate`](https://rstudio.github.io/reticulate/):
+
+```r
+library(reticulate)
+datasets <- import("philanthropy.datasets")
+models   <- import("philanthropy.models")
+
+df <- datasets$generate_synthetic_donor_data(n_samples = 1000L, random_state = 0L)
+X  <- as.matrix(df[c("total_gift_amount", "years_active", "event_attendance_count")])
+y  <- df$is_major_donor
+
+model  <- models$DonorPropensityModel(n_estimators = 100L, random_state = 0L)
+model$fit(X, y)
+scores <- model$predict_affinity_score(X)   # 0-100 affinity scores
+```
+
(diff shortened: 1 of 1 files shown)


shivamlalakiya:
Superseded: #132 (merged) carries this exact reticulate quickstart section forward as its top commit, so it's already in main. Thanks for the writeup, the L-suffix callout was the right catch.

PR #135, opened 2026-08-21
fix: CRMCleaner silently corrupted complex amounts into wrong finite floats (#129)

### LINKED ISSUE #129: CRMCleaner.transform silently corrupts complex amounts into wrong finite floats
Found while writing the coverage tests for #118 (this is the observation I flagged in that PR body). Reproducible on current main; filing as its own issue per the request in #126 for genuine external-user signal — this is a real friction point I hit, not manufactured.

## What happens

A DataFrame whose amount column mixes regular values with one Python `complex` slips through every guard in `CRMCleaner.transform` and produces a **plausible-looking but wrong finite number**, instead of raising or NaN-ing:

```python
import numpy as np
import pandas as pd
from philanthropy.preprocessing import CRMCleaner

X_fit = pd.DataFrame({
    "gift_date": ["2023-01-01", "2023-02-01"],
    "gift_amount": [5.0, 10.0],
})
t = CRMCleaner().fit(X_fit)

X_bad = pd.DataFrame({
    "gift_date": ["2023-03-01"],
    "gift_amount": pd.Series([np.complex128(3 + 4j)], dtype=object),
})
print(t.transform(X_bad))
# gift_date              2023-03-01 00:00:00
# gift_amount                          -34.0      <- silently corrupted
```

`3+4j` becomes **-34.0**. For context on how easy this is to hit: `np.complex128` values show up in real CRM exports when a formula column round-trips through Excel/openpyxl.

## Root cause (three guards miss it in sequence)

1. The mixed str/complex frame raises `DTypePromotionError` in the first `validate_data` pass; the object-cast retry introduced for exactly this kind of input lets it through.
2. `np.iscomplexobj(X_arr)` returns **False** because after the retry the array has object dtype — `iscomplexobj` inspects dtype, not elements.
3. `_coerce_currency_to_float` takes the string path: `str(3+4j)` is `"(3+4j)"`, which matches the parenthesised-negative rule (`-\1`), then character-stripping deletes the parens and `j`, leaving `"-34"` → `-34.0`. Sign flipped, imaginary part dropped, digits concatenated.

The same frame passed directly to `validate_data` raises `Complex data not supported` cleanly at fit time — the corruption is specific to the transform-time retry path.

## Suggested direction

In `_coerce_currency_to_float`'s string path, reject parsed-out-of-nothing values that were originally complex before the regex cleanup, or check element-wise `isinstance(v, complex)` on object columns before casting to str. A regression test would assert either a raise or NaN for the complex cell — anything but a finite number. Happy to send the PR if you tell me which behavior you want (raise vs NaN); my instinct is NaN-plus-warning, consistent with "values that still don't parse become NaN".

slegarraga:
Closes #129

## What happened (reproduced on current main)

`CRMCleaner.transform` turned `3+4j` into **-34.0** — sign flipped, imaginary part dropped, digits concatenated — via three guards missing in sequence (full root-cause chain in #129).

## The fix (two layers)

1. **Element-wise complex guard** in `_coerce_currency_to_float`: on object columns, cells that are actually `complex` are masked to NaN with a `UserWarning` *before* the string cleanup can mangle them. NaN + warning follows the documented "values that still don't parse become NaN" contract.
2. **Contract preservation**: `had_value` is now captured *before* masking, so a column where nothing parses (all-complex included) still raises `could not parse`, exactly as documented.

## Verification (executed)

- The exact repro from #129 now returns `NaN` + one `UserWarning mentioning "complex"` instead of `-34.0`.
- Two new regression tests cover both contracts (mixed-column → NaN+warn; all-unparseable → raise).
- Full suite: **1843 passed / 23 skipped / 0 failed**.

## Notes for review

- No public API change; behavior only differs where it was previously silently wrong.
- Happy to switch to hard-raising on any complex cell if you prefer strictness over tolerance — one-line change + test tweak.


shivamlalakiya:
Good root-cause fix, well tested (the whole-column-still-raises case especially). One gap before merge: AGENTS.md requires a CHANGELOG.md entry under `## [Unreleased]` for every PR touching `philanthropy/`, and this one doesn't have one yet (you're already credited in CONTRIBUTORS.md). Could you add a line for the complex-value fix? Holding the merge until then.

slegarraga:
Done in 384f2ba: added the `## [Unreleased]` entry under `### Fixed`,
describing both layers (complex cells masked to NaN with a warning naming
them; all-unparseable columns still raise per the documented contract).
Thanks for taking a close look at the whole-column test.

slegarraga:
Heads up: main moved (congrats on landing #140!) and this branch picked up a stale-conflict flag, so I merged current `main` in and re-ran the full local gates: `make ci` green (1904 passed, 97.06% coverage) and `make riskcov` green (96%). The CHANGELOG now carries both our `### Fixed` entry and the new `### Added` section side by side, no manual untangling needed. PR shows conflict-free again.

PR #139, opened 2026-08-22
test: RFMTransformer.fit never exercises the ndarray (non-Da

Mr-Neutr0n:
Fixes #49

Added `test_rfm_transformer_ndarray_input_raises_missing_columns` to cover `RFMTransformer.fit`’s bare-ndarray path and verify it raises `ValueError` for missing required columns.

Local tests pass.

---
This change was prepared with AI assistance under human direction and review.

Issue #82, opened 2026-08-20
Add a second person who can merge, or change the rule that says you cannot


shivamlalakiya:
## The problem

`.github/CODEOWNERS` is `* @shivamlalakiya`. Org members: `shivamlalakiya`. Collaborators: `shivamlalakiya`. `AGENTS.md` says:

> never merge your own PR; open it and leave the merge to review.

Exactly one person can merge, and that person is not allowed to merge their own work. **The rule as written cannot hold**, and the repository visibly does not follow it. A reviewer who reads `AGENTS.md` and then the merge history sees the contradiction.

## Why it matters more than it looks

It is the binding constraint on everything else. With one merger:

- Green PRs sit. PR #30 sat 40 hours; neither of the two contributors who filed it returned.
- Agent-fleet operators optimise merged-PR throughput and deprioritise slow repos, so slow merges repel the next contributor harder than having no open issues at all.
- If the maintainer goes quiet, nobody on earth can merge a green PR and the one working contribution channel becomes a stale-PR graveyard.

## What would resolve it

Either of:

1. **Add a second org owner** whose only job is to click merge on a green PR. Then state the target publicly, e.g. "green PRs merged within 24h".
2. **Amend `AGENTS.md` to describe what actually happens** (maintainer merges after CI is green; a second reviewer is required when one is available).

Option 1 is better. Option 2 is still better than the status quo, because a rule the repo keeps beats a stricter rule it visibly breaks.

This needs a person, so it cannot be done by a PR.

PR #144, opened 2026-08-23
test: cover FinancialForecastModel linear-only predict and short-history AR padding

### LINKED ISSUE #59: test: `FinancialForecastModel` has two untested branches — linear-only `predict` and short-history AR padding
## Summary

`FinancialForecastModel` has two branches that no test in
`tests/test_forecast_model.py` exercises:

1. **`predict` — linear-only path.** `predict` adds
   `self.nonlinear_model_.predict(X_imp)` to the linear prediction only
   `if self.nonlinear_model_ is not None`. Every existing test fits with the
   hybrid (linear + `MLPRegressor` residual) configuration on, so the
   `nonlinear_model_ is None` branch — i.e. `predict` returning the bare linear
   prediction — never runs.
2. **`predict_revenue_forecast` — short-history padding.** When the AR window
   `p` (`self.ar_order`) exceeds the length of `history`, the seed window is
   padded with `self.y_mean_`: `window = list(history[::-1]) + [self.y_mean_] *
   (p - history.size)`. Every existing forecast test supplies at least `p`
   periods of history, so the padding branch never runs.

### Where

| File | Locate with |
|---|---|
| `philanthropy/models/_forecast.py` | `grep -n "if self.nonlinear_model_ is not None" philanthropy/models/_forecast.py` |
| `philanthropy/models/_forecast.py` | `grep -n "p - history.size" philanthropy/models/_forecast.py` |

### What to change

1. Open `tests/test_forecast_model.py` and find the existing `predict` test
   (locate with `grep -n "def test.*predict" tests/test_forecast_model.py`).
   Add a sibling that fits `FinancialForecastModel` with parameters that leave
   `nonlinear_model_` as `None` (check the constructor — likely a
   `use_nonlinear=False` or equivalent flag; grep `def __init__` in
   `_forecast.py` to confirm the exact parameter name), then asserts `predict`
   matches `linear_model_.predict` on imputed input directly.
2. Add a second test that calls `predict_revenue_forecast` with `X` containing
   fewer rows than `ar_order`, and asserts it returns a finite `(horizon,)`
   array without raising — the interesting behaviour is that the padded window
   uses `y_mean_` rather than crashing on a short window.

### Tests to add or extend

- File: `tests/test_forecast_model.py`
- Add: `test_predict_without_nonlinear_model_uses_linear_only`
- Add: `test_predict_revenue_forecast_pads_short_history_with_training_mean`

### Done when

```bash
python -m pytest tests/test_forecast_model.py --cov=philanthropy.models._forecast --cov-report=term-missing -q
```

`philanthropy/models/_forecast.py`'s `Missing` column no longer lists either
guard above. (`_forecast.py` may still show other unrelated partial branches —
leave those alone; they are out of scope here.)

slegarraga:
Closes #59. Test-only, no source change.

## What

Two named tests in `tests/test_forecast_model.py`, exactly as specced:

1. `test_predict_without_nonlinear_model_uses_linear_only`: fits on a single training sample, which leaves the residual network unfitted (`nonlinear_model_ is None`), then asserts `predict(X)` matches `linear_model_.predict` on the same NaN-free input exactly. The probe is NaN-free on purpose, so imputation is the identity and the comparison against the linear sub-model is direct.
2. `test_predict_revenue_forecast_pads_short_history_with_training_mean`: fits with `ar_order=5`, calls `predict_revenue_forecast` with 2 context rows (fewer than `p`), asserts a finite `(horizon,)` array, and then reproduces the documented roll-forward by hand from the frozen fitted statistics (`predict` history reversed most-recent-first, padded to `p` with `y_mean_`), so the test pins the pad value rather than just finiteness.

## Evidence

Coverage of `philanthropy/models/_forecast.py` from this file, before and after (branch coverage):

```
before: 90 stmts, Miss 1, BrPart 2, 97%, Missing: 288->290, 341
after:  90 stmts, Miss 0, BrPart 0, 100%
```

`341` is the `window = list(history[::-1]) + [self.y_mean_] * (p - history.size)` padding line and `288->290` the `if self.nonlinear_model_ is not None` guard, the two branches named in the issue. Unrelated partial branches elsewhere in the module: none left in this file, nothing else touched.

## Local verification

- `python -m pytest tests/ -q`: 1910 passed, 25 skipped.
- `python -m pytest tests/test_forecast_model.py -q`: 78 passed, 1 skipped.
- `flake8 tests/test_forecast_model.py`: clean.

Test-only PR, so no CHANGELOG entry per the precedent noted in #139.

### DIFF
--- a/tests/test_forecast_model.py
+++ b/tests/test_forecast_model.py
@@ -91,6 +91,22 @@ def test_hybrid_uses_nonlinear_component(revenue_Xy):
     assert not np.allclose(linear_only, hybrid)
 
 
+def test_predict_without_nonlinear_model_uses_linear_only(revenue_Xy):
+    """A single training sample leaves the residual network unfitted
+    (`nonlinear_model_ is None`): `predict` must return the bare linear
+    prediction instead of adding any residual term."""
+    X, y = revenue_Xy
+    model = FinancialForecastModel(random_state=0).fit(X[:1], y[:1])
+    assert model.nonlinear_model_ is None
+    # NaN-free probe, so imputation is the identity and comparing against
+    # `linear_model_.predict` directly is exact.
+    X_probe = np.asarray(X[:3], dtype=float)
+    np.testing.assert_allclose(
+        model.predict(X_probe),
+        model.linear_model_.predict(X_probe),
+    )
+
+
 # ---------------------------------------------------------------------------
 # predict_revenue_forecast
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


shivamlalakiya:
@slegarraga this isn't a thank-you note, it's an offer.

Eighteen of the thirty-three external pull requests merged into this project are yours. #135 caught complex values silently corrupting into wrong finite floats, which is the worst class of bug this library can have because it ships a plausible number rather than an error. #123 closed a network-scheme path at the read boundary. #128 is the CI job that now blocks any pull request skipping changelog or contributor credit, including mine. And in #119 you credited @shubhrai23 in CONTRIBUTORS.md without anyone asking you to.

The project has a bus factor of 1. It's tracked in #82, it's the weakest thing about the repository, and it's not an engineering problem I can fix by myself. I merge my own pull requests because no second person holds merge rights, and I have to disclose exactly that in the submission notes when this goes to JOSS next month.

So: would you take commit and merge rights?

Concretely what that means, and what it does not:

- You could review and merge pull requests that aren't your own, once CI is green.
- No obligation to any volume, any response time, or any area of the codebase.
- A standing right to say "not looking at this one" with no explanation.
- No obligation to keep doing it. If you want out later, you say so and that's the whole process.

If you'd rather not, that's a completely fine answer, it needs no reason, and it changes nothing about the standing invitation to keep contributing however you like. Either way, I'd rather ask than assume.

PR #146, opened 2026-08-24
test: cover DataFrame feature names in MovesManagementClassifier

### LINKED ISSUE #54: test: `MovesManagementClassifier.fit` never exercises the DataFrame `feature_names_in_` branch
### What's wrong

`MovesManagementClassifier.fit` branches on whether `X` has a `.columns`
attribute, setting `feature_names_in_` from the column names when it does.
Every test in `tests/test_moves.py` fits with a bare `ndarray`, so the
`pd.DataFrame` branch never runs and `feature_names_in_` is never exercised via
a real DataFrame input.

### Where

| File | Locate with |
|---|---|
| `philanthropy/models/_moves.py` | `grep -n 'hasattr(X, "columns")' philanthropy/models/_moves.py` |

### What to change

1. Open `tests/test_moves.py` and find the existing `fit`-related fixture/test
   (locate with `grep -n "def test_" tests/test_moves.py | head`) for the style
   used to build `X`/`y`.
2. Add a test that fits `MovesManagementClassifier` on a `pandas.DataFrame` with
   named columns and asserts `feature_names_in_` equals those column names
   (`np.array([...], dtype=object)`).

### Tests to add or extend

- File: `tests/test_moves.py`
- Add: `test_fit_with_dataframe_sets_feature_names_in`

### Done when

```bash
python -m pytest tests/test_moves.py --cov=philanthropy.models._moves --cov-report=term-missing -q
```

`philanthropy/models/_moves.py` reads 100% covered with an empty `Missing` column.

stoppo22:
## What & why

Adds regression coverage for `MovesManagementClassifier.fit` when the feature matrix is provided as a `pandas.DataFrame`.

The new test verifies that fitting on named DataFrame columns correctly preserves those names in `feature_names_in_`.

Closes #54.

Local CI completed successfully:

```bash
uv run make ci
```

Result:
- 1909 passed
- 25 skipped
- 97.09% total coverage

## Checklist
- [x] `make ci` passes locally (lint → collection → tests → coverage ≥ 92%)
- [ ] New/changed public API has docstrings and is exported in the subpackage `__init__.py`
- [x] Tests added or updated
- [x] `CHANGELOG.md` updated under `[Unreleased]`
- [x] Added yourself to `CONTRIBUTORS.md` (skip if you would rather not be listed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   NaN with a `UserWarning` naming them, and a column where nothing parses
   (all-complex included) still raises `could not parse` per the documented
   contract. Closes #129.
+  - Added regression coverage ensuring `MovesManagementClassifier.fit` preserves
+  DataFrame column names in `feature_names_in_`. Closes #54.
 ### Added
 - **`models.GiftIntervalCalibrator`**: distribution-free intervals on a dollar
   amount. Wraps an already-fitted regressor (`AskAmountRecommender`,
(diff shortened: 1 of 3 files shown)


shivamlalakiya:
@stoppo22 thanks again for #146. The DataFrame feature-name path in `MovesManagementClassifier` is exactly the case real users hit, because nobody outside a test suite passes a bare ndarray, and it's exactly the case synthetic tests miss for the same reason.

PR #64, opened 2026-08-11
docs: add missing docstrings to UpliftTLearner and cli.main

shubhrai23:
Resolves #58

**What changed:** Added comprehensive NumPy-style docstrings to the `UpliftTLearner.predict` method and the `cli.main` function.
**Why:** To ensure the API documentation is complete and consistent with the rest of the module, making it easier for developers to understand the required arguments and return types.

PR #147, opened 2026-08-26
DEP: warn when drop_repeat_donors default is used (closes #108)

### LINKED ISSUE #108: FiscalYearGroupedSplitter: drop_repeat_donors defaults to False, so the default path still leaks a static per-donor label across folds
### What's wrong

`FiscalYearGroupedSplitter(drop_repeat_donors=...)` defaults to `False`. For a
static per-donor label (e.g. `is_major_donor`, the label used in the README,
the benchmarks page, and `scripts/benchmark_models.py`), leaving it at the
default lets the same donor's rows land in both the train and test fold of a
split. That is the exact leakage this splitter exists to prevent, and it is
live in the default path, not an opt-in footgun.

### Where

| File | Locate with |
|---|---|
| `philanthropy/model_selection/_temporal_donor_splitter.py` | `grep -n "drop_repeat_donors: bool = False" philanthropy/model_selection/_temporal_donor_splitter.py` |

### What to change

1. Add a `DeprecationWarning` in `FiscalYearGroupedSplitter.split` (or
   `__init__`) when `drop_repeat_donors` is left at its default `False`,
   stating that the default will become `True` in a future release and that
   callers relying on repeat donors appearing in both folds should pass
   `drop_repeat_donors=False` explicitly to silence it.
2. Update the class docstring's `drop_repeat_donors` parameter entry to state
   the planned default flip and the release it lands in, following the same
   pattern already used for `WealthScreeningImputerKNN(group_col_idx=...)`
   (see `CHANGELOG.md`, `## [Unreleased]` section, `Deprecated`).
3. Add a `## [Unreleased]` `CHANGELOG.md` entry and a `CONTRIBUTORS.md` line
   in the same PR (required by `AGENTS.md`).

### Tests to add or extend

- File: `tests/test_temporal_donor_splitter.py`
- Add: `test_default_drop_repeat_donors_emits_deprecation_warning`
- Add: `test_explicit_drop_repeat_donors_false_silences_warning`

### Done when

```bash
python -c "
import warnings
from philanthropy.model_selection import FiscalYearGroupedSplitter
with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter('always')
    FiscalYearGroupedSplitter()
    assert any(issubclass(x.category, DeprecationWarning) for x in w)
"
make ci
```

### First time here?

Read [CONTRIBUTING.md](../../CONTRIBUTING.md) and [AGENTS.md](../../AGENTS.md).
Add a `## [Unreleased]` CHANGELOG entry and yourself to `CONTRIBUTORS.md` in
the same PR.

shubhrai23:
## Summary

`FiscalYearGroupedSplitter(drop_repeat_donors=...)` currently defaults to `False`, which silently allows repeat donors across train and test folds. This PR deprecates that default so it will change to `True` in 0.8.0, following the pattern established by `WealthScreeningImputerKNN.group_col_idx`.

### `philanthropy/model_selection/_temporal_donor_splitter.py`
- Default changed from `False` to the sentinel `"warn"`.
- `__init__` now emits a `DeprecationWarning` when the default is left untouched.
- `split()` and `get_n_splits()` resolve `"warn"` to `False` internally, so behaviour is unchanged.
- Docstring updated with a `.. deprecated:: 0.7.0` block.
- Doctests updated to pass `drop_repeat_donors=False` explicitly to avoid warnings.

### `tests/test_model_selection.py`
- Added `test_default_drop_repeat_donors_emits_deprecation_warning`.
- Added `test_explicit_drop_repeat_donors_false_silences_warning`.
- All existing tests now pass `drop_repeat_donors=False` explicitly or wrap in `pytest.warns(DeprecationWarning)`.

### `tests/test_deprecations.py`
- Added `FiscalYearGroupedSplitter.drop_repeat_donors` to the `DEPRECATIONS` registry so the meta-test passes.

### `tests/test_leakage.py`
- Silenced the new warning by passing `drop_repeat_donors=False` explicitly.

### `CHANGELOG.md`
- Added entry under `[Unreleased] / Deprecated`.

## Checklist
- [x] Changelog updated
- [x] Author credited in CONTRIBUTORS.md (already listed from prior PRs)
- [x] All tests pass locally (64 passed)
- [x] flake8 clean
- [x] mypy clean

Closes #108

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
+### Deprecated
+- `FiscalYearGroupedSplitter`'s default for `drop_repeat_donors` (currently `False`) is deprecated and will change to `True` in 0.8.0. Leaving it at its default now emits a `DeprecationWarning`. Pass `drop_repeat_donors=False` explicitly to silence the warning and retain current behavior. Closes #108, by @shubhrai23.
+
 ### Added
 - `credit-guard` CI job: pull requests touching `philanthropy/` must also
   update this changelog, and the author must be credited in
--- a/philanthropy/model_selection/_temporal_donor_splitter.py
+++ b/philanthropy/model_selection/_temporal_donor_splitter.py
@@ -24,7 +24,7 @@ Typical usage
 >>> from philanthropy.model_selection import FiscalYearGroupedSplitter
 >>> X = np.zeros((100, 3))
 >>> fiscal_years = np.array([2019]*20 + [2020]*30 + [2021]*25 + [2022]*25)
->>> splitter = FiscalYearGroupedSplitter(n_splits=3)
+>>> splitter = FiscalYearGroupedSplitter(n_splits=3, drop_repeat_donors=False)
 >>> splits = list(splitter.split(X, groups=fiscal_years))
 >>> len(splits)
 3
@@ -72,6 +72,12 @@ class FiscalYearGroupedSplitter(BaseCrossValidator):
         training (useful when gift officers use current-year pipeline
         intelligence that would not have been available historically).
     drop_repeat_donors : bool, default=False
+        .. deprecated:: 0.7.0
+            Leaving ``drop_repeat_donors`` at its default emits a
+            ``DeprecationWarning``. The default changes to ``True`` in 0.8.0.
+            Pass ``drop_repeat_donors=False`` explicitly to silence this warning
+            and keep the current behaviour.
+
         Whether to remove from each test fold any donor who already appears in
         that fold's training rows.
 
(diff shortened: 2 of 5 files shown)


shivamlalakiya:
@shubhrai23 four merged pull requests here and the arc is unusual. You started with docstrings in #64 and #65, went to test coverage in #66, and ended up shipping #147, an actual deprecation warning on a default parameter. That last one is a public API change with a compatibility promise attached, not a starter task, and it's the one I'd point at if someone asked what a good-first-issue funnel is supposed to produce.

Two things.

First, I'm asking the people who've contributed here what was annoying about the setup, since you all did it recently enough to still remember where the instructions were wrong: https://github.com/PhilanthroPy-Project/PhilanthroPy/discussions/188

Second, if there's an open issue you'd want to take a bigger swing at, say which one and I'll leave it assigned to you rather than letting it get picked up by someone else. Given #147, I don't think the `good first issue` label is where your ceiling is.

shubhrai23:
Thanks for the encouragement, can I grab issue #155 ?

I took a brief look and it seems like a solid clean-up. I plan to extract that repeated validate_data try-catch block into a single helper function to dry things up. I also realized why that complex-data guard is unreachable , np.iscomplexobj() evaluates to False on the mixed-type object arrays that validate_data returns, so that check is dead code and can just be removed.

PR #148, opened 2026-09-01
test: `PlannedGivingIntentScorer.predict_intent_score` never

### LINKED ISSUE #56: test: `PlannedGivingIntentScorer.predict_intent_score` never exercises its single-class fallback
### What's wrong

`PlannedGivingIntentScorer.predict_intent_score` short-circuits to
`np.zeros(proba.shape[0])` when `predict_proba` returns fewer than two columns
(the single-class-seen-at-fit case). This is the same shape of guard as the one
fixed for `PropensityScorer.predict_proba` in #50, but in a different class and
method — nothing in `tests/test_planned_giving.py` fits on single-label data, so
this branch has never run.

### Where

| File | Locate with |
|---|---|
| `philanthropy/models/_planned_giving.py` | `grep -n "proba.shape\[1\] < 2" philanthropy/models/_planned_giving.py` |

### What to change

1. Open `tests/test_planned_giving.py` and look at the existing
   `predict_intent_score` test (locate with
   `grep -n "def test.*intent_score" tests/test_planned_giving.py`) for fixture
   style.
2. Add a test that fits `PlannedGivingIntentScorer` on a `y` containing only one
   label, then calls `predict_intent_score` and asserts every score is `0.0`.

### Tests to add or extend

- File: `tests/test_planned_giving.py`
- Add: `test_predict_intent_score_single_class_returns_zeros`

### Done when

```bash
python -m pytest tests/test_planned_giving.py --cov=philanthropy.models._planned_giving --cov-report=term-missing -q
```

`philanthropy/models/_planned_giving.py` reads 100% covered with an empty
`Missing` column.

Mr-Neutr0n:
Fixes #56

Added `test_predict_intent_score_single_class_returns_zeros` to `tests/test_planned_giving.py` to cover the `predict_intent_score` branch that returns all zeros when `predict_proba` has fewer than two columns; no source changes. (The issue’s coverage command could not be run here because `pytest-cov` triggers a numpy import error in this environment; all 16 tests in the file pass without coverage.)

Local tests pass.

---
This change was prepared with AI assistance under human direction and review.

### DIFF
--- a/tests/test_planned_giving.py
+++ b/tests/test_planned_giving.py
@@ -12,6 +12,7 @@ import pytest
 from sklearn.base import clone
 from sklearn.exceptions import NotFittedError
 
+from philanthropy.models import PlannedGivingIntentScorer
 from philanthropy.preprocessing import PlannedGivingSignalTransformer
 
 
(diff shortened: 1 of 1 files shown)


shivamlalakiya:
Closing as duplicate: #149 covers the same issue (#56) with the same core test, plus extra coverage for the multi-class branch and `predict`. Thanks for the PR, hari, and for calling out the coverage tooling issue in the description.

PR #149, opened 2026-09-01
test: cover planned giving single-class intent score

### LINKED ISSUE #56: test: `PlannedGivingIntentScorer.predict_intent_score` never exercises its single-class fallback

Mohd-Hamza-Khan:
Summary

Adds test coverage for the single-class probability fallback in PlannedGivingIntentScorer.predict_intent_score.

What changed
Added test_predict_intent_score_single_class_returns_zeros.
Simulates a single-class predict_proba() output with one probability column.
Verifies that predict_intent_score() returns a zero score for every sample.
No production code was changed.
Why

The existing single-class fallback was not covered by the test suite. This test exercises the proba.shape[1] < 2 branch and verifies that it returns zeros as expected.

Testing
python -m pytest tests/test_planned_giving.py -q
python -m pytest tests/test_planned_giving.py --cov=philanthropy.models._planned_giving --cov-report=term-missing -q
make ci
make riskcov

All checks pass.

Resolves #56

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,9 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   flips between the three is a ranking of the tail), median width, and
   `width_ratio` = median width over median target. A valid interval can carry no
   information; the ratio is what separates the two.
+- Test coverage for `PlannedGivingIntentScorer.predict_intent_score`: the
+  single-class `predict_proba` fallback path that returns an all-zero score.
+  (#56)
 
 ### Changed
 - `ShareOfWalletScorer` output column 0 is renamed `sow_score` →
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -41,6 +41,10 @@ contribution. Code, docs, tests, and review all count.
 - [@stoppo22](https://github.com/stoppo22): added DataFrame feature-name
   coverage for `MovesManagementClassifier.fit`
   ([#146](https://github.com/PhilanthroPy-Project/PhilanthroPy/pull/146)).
+- [@Mohd-Hamza-Khan](https://github.com/Mohd-Hamza-Khan): added the missing
+  single-class fallback coverage for
+  `PlannedGivingIntentScorer.predict_intent_score`
+  ([#149](https://github.com/PhilanthroPy-Project/PhilanthroPy/pull/149)).
 
 ## Getting listed
 
--- a/tests/test_planned_giving.py
+++ b/tests/test_planned_giving.py
@@ -13,7 +13,8 @@ from sklearn.base import clone
 from sklearn.exceptions import NotFittedError
 
 from philanthropy.preprocessing import PlannedGivingSignalTransformer
-
+from philanthropy.models import PlannedGivingIntentScorer
+from unittest.mock import patch
 
 # ---------------------------------------------------------------------------
 # Shared helper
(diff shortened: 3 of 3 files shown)


shivamlalakiya:
@Mohd-Hamza-Khan thanks again for #149. The single-class intent-score case is a real edge rather than a hypothetical one: a small planned-giving cohort where every donor falls on one side is exactly the shape of data a mid-size shop actually has, and it's the shape that makes a classifier fall over.

Mohd-Hamza-Khan:
@shivamlalakiya Thanks for asking! I really appreciate the detailed contribution guide. The most annoying part for me was running the test/setup commands on Windows. The CONTRIBUTING.md instructions are mostly Unix/Linux-oriented (make ci, sh scripts/install_hooks.sh, etc.), so it wasn't immediately clear what the equivalent workflow should be.

It would be really helpful to include Windows-specific instructions, such as using nmake where appropriate (or another recommended alternative to make), how to run the hook script on Windows, and any other Windows-compatible commands.

shivamlalakiya:
@Mohd-Hamza-Khan this is exactly the answer I was hoping to get, and it is the second independent report of the same thing, which is what moved it from "someone's local setup" to a real gap. @AsavariCharati hit it too and wrote it up in #215, which just merged, so `CONTRIBUTING.md` now has a "Testing on Windows" section and `docs/how-to/develop_and_test.md` has the `pytest-cov` note. Worth a look to check it matches what you actually ran into; if it misses something, say so on #195 or open a new issue and I will fix it.

On `nmake` specifically: I looked at it and decided against recommending it. It is not a drop-in for GNU `make` here. It ships with the MSVC build tools rather than with Python or Git, its syntax differs enough that `Makefile` would need a parallel `Makefile.win` to work under it, and that second file is then free to drift from the first without anything catching it. Since the `Makefile` targets are only shell wrappers around `python -m flake8`, `python -m mypy`, and `python -m pytest`, documenting those commands directly costs nothing and cannot drift out of sync with a tool nobody on Windows has installed. That is the route #215 took.

The hook question you raised has a worse answer than "how do I run it." `sh scripts/install_hooks.sh` needs Git Bash to install at all, and even once installed the pre-push hook does not fire when you push from plain PowerShell. Confirmed by testing, not assumed. So a Windows contributor pushing from PowerShell gets no local gate whatsoever and no warning that it was skipped. That is now written down, with the manual command list as the substitute.

One more thing your Windows work is upstream of. Running the suite on Windows surfaced 11 failures that turned out to be a genuine bug rather than an environment artifact: `ensure_local_path` calls `urlparse` and reads the drive letter in `C:\data\gifts.csv` as a URI scheme `'c'`, so it rejected every absolute Windows path across four public entry points. Filed as #217 and fixed in #218, merged a few minutes ago. Nobody would have found that without someone running the gate on Windows.

Thank you for taking the time to answer a question you got nothing out of answering.

PR #174, opened 2026-09-06
test: add edge-case tests for WealthPercentileTransformer missing columns (#169)

### LINKED ISSUE #169: test: `WealthPercentileTransformer` returns all-NaN ranks for an all-missing wealth column, and nothing checks it
### What's wrong

If every value in a wealth column is missing in the training data,
`WealthPercentileTransformer.fit` stores an empty reference array for it and
`transform` returns `NaN` for that column's percentile rank rather than dividing by
zero. That is the right behaviour and it is not tested, so the branch is one edit away
from silently becoming a `RuntimeWarning` and a column of `inf`.

This is not hypothetical for the domain. A purchased wealth screen that matched none
of your records comes back as exactly this: a column that exists, is the right length,
and is entirely null.

Current behaviour, which the test should pin:

```python
df = pd.DataFrame({"net_worth": [np.nan, np.nan, np.nan], "other": [1.0, 2.0, 3.0]})
t = WealthPercentileTransformer().fit(df)
t.percentile_lookup_["net_worth"]     # array([], dtype=float64)
t.transform(df)
# [[nan  1. nan]
#  [nan  2. nan]
#  [nan  3. nan]]
t.get_feature_names_out()             # ['net_worth' 'other' 'net_worth_pct_rank']
```

Note the third column is still emitted and still named. The transformer does not
quietly drop it, which matters: a pipeline downstream of this has a stable output
width whether or not the screen matched anything.

### Where

| File | Locate with |
|---|---|
| the untested guard | `grep -n 'len(ref) == 0' philanthropy/preprocessing/_wealth_percentile.py` |
| where the reference array is frozen | `grep -n 'percentile_lookup_\[col\]' philanthropy/preprocessing/_wealth_percentile.py` |
| the existing tests for this class | `grep -n 'WealthPercentileTransformer' tests/test_preprocessing.py` |

### What to change

No source change. This is a missing test.

### Tests to add or extend

- Add: `test_wealth_percentile_all_missing_column_yields_nan_ranks`

  1. Build a DataFrame with one wealth column that is entirely `NaN` and one ordinary
     numeric column.
  2. Fit, and assert `percentile_lookup_["net_worth"].size == 0`.
  3. Transform, and assert the appended rank column is entirely `NaN` while the
     ordinary column passes through unchanged.
  4. Assert `get_feature_names_out()` still names the rank column, so the output width
     does not depend on whether the screen matched.
  5. Assert no warning is raised: wrap the `transform` call in
     `warnings.catch_warnings()` with `simplefilter("error")`. A future
     divide-by-zero would surface here rather than as `inf` in someone's model.

- Add: `test_wealth_percentile_partially_missing_column_ranks_only_observed_values`

  The neighbouring case, so the first test cannot pass for the wrong reason: with some
  values present, the rank column is `NaN` exactly where the input was `NaN` and a
  number everywhere else.

### Done when

```bash
# Exits 0 only once both tests exist and pass.
python -m pytest tests/test_preprocessing.py -k "wealth_percentile and missing" -q
test "$(grep -c 'def test_wealth_percentile_all_missing_column_yields_nan_ranks' tests/test_preprocessing.py)" = "1"
make ci
make riskcov
```

### First time here?

Read [CONTRIBUTING.md](../../CONTRIBUTING.md) and [AGENTS.md](../../AGENTS.md). This
one needs no source change, so it is a good first landing. Add a `## [Unreleased]`
CHANGELOG entry and yourself to `CONTRIBUTORS.md` in the same PR.

Larslllllll:
## Summary

Add two unit tests to `tests/test_preprocessing.py` that pin the behaviour of
`WealthPercentileTransformer` when wealth columns contain missing values.

## Changes

- **`tests/test_preprocessing.py`** — New `TestWealthPercentileTransformer` class with:
  - `test_all_missing_column_yields_nan_ranks`: Pins the all-NaN column branch:
    the guard stores an empty reference array, `transform` returns NaN ranks,
    output width is stable, and no warning is raised.
  - `test_partially_missing_column_ranks_only_observed_values`: Pins the
    partially-NaN column branch: observed values get numeric ranks, NaN input
    rows get NaN ranks.
- **`CHANGELOG.md`** — Added entry under `[Unreleased] > Added`.
- **`CONTRIBUTORS.md`** — Added self.

## Closes

Closes #169.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   project's standing rules.
 
 ### Added
+- Tests pinning the two untested branches of `WealthPercentileTransformer`:
+  the all-missing column path (returns NaN ranks, keeps a stable output
+  width, raises no warning) and the partially-missing column path (NaN
+  input rows get NaN rank, observed rows get numeric rank). Closes #169.
 - **`datasets.make_donor_panel`** (Tier 2, Beta): a seeded multi-year donor
   panel returning gift-level rows rather than one aggregated row per donor.
   `generate_synthetic_donor_data` cannot demonstrate `RFMTransformer` (needs a
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -45,6 +45,10 @@ contribution. Code, docs, tests, and review all count.
   single-class fallback coverage for
   `PlannedGivingIntentScorer.predict_intent_score`
   ([#149](https://github.com/PhilanthroPy-Project/PhilanthroPy/pull/149)).
+- **Lars** ([@Larslllllll](https://github.com/Larslllllll)): added the missing
+  unit-test coverage for `WealthPercentileTransformer`'s all-missing and
+  partially-missing column branches in `WealthPercentileTransformer` (closes
+  [#169](https://github.com/PhilanthroPy-Project/PhilanthroPy/issues/169)).
 
 ## Getting listed
 
--- a/tests/test_preprocessing.py
+++ b/tests/test_preprocessing.py
@@ -32,6 +32,7 @@ from philanthropy.preprocessing import (
     EncounterTransformer,
     FiscalYearTransformer,
     WealthScreeningImputer,
+    WealthPercentileTransformer,
 )
 
 # Property-based tests for these transformers live in tests/test_properties.py.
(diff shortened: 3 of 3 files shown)


shivamlalakiya:
@Larslllllll thanks again for #174. Your edge-case tests on `WealthPercentileTransformer` are what set up the fix that landed in #179 a couple of days later: the tests pinned down what the missing-column behaviour actually was, which is what made it obvious that the behaviour was wrong.

I also owe you a note about #176. You and another contributor solved the same four-line problem on the same evening and I closed one of them, which cost somebody an evening for nothing. That was the project's process failure, not yours: there was no way to claim an issue before opening a pull request. #187 is merged now and adds a claim-before-PR rule to CONTRIBUTING.md, so issues get assigned before anyone writes code.

Which leads to the question I'm asking everyone who's contributed here. When you set the repo up, what was the most annoying part? Install, the test suite, working out where the code lived, the CI gates, the duplicate-PR mess above, anything at all. There's a thread for it here: https://github.com/PhilanthroPy-Project/PhilanthroPy/discussions/188, or open an issue if it's concrete enough to fix.

Issue #156, opened 2026-09-05
`WealthPercentileTransformer` silently ignores a `wealth_cols` list that matches no column


shivamlalakiya:
### What's wrong

`WealthPercentileTransformer(wealth_cols=[...])` silently does nothing when none
of the named columns exist in the training frame.

`fit` resolves `wealth_cols` against `feature_names_in_`, keeps only the ones it
finds, and if that leaves an empty list it hits this:

```python
if len(self.imputed_cols_) == 0 and self.wealth_cols is not None:
     # If we expected columns but found none, we should probably warn or raise
     pass
```

The comment is the whole implementation. A gift officer who passes
`wealth_cols=["estimated_networth"]` (one typo away from
`estimated_net_worth`) gets a fitted transformer, a clean `transform`, no
percentile columns, no warning, and a model that quietly trained without the
wealth signal it was supposed to have. Silence is the wrong answer here: the
caller named specific columns, so naming ones that do not exist is a mistake, not
a preference.

There is a second, smaller defect a few lines above, in the same method:

```python
X = validate_data(self, X, ensure_all_finite="allow-nan", reset=True)

if hasattr(X, "columns"):
     self.feature_names_in_ = np.array(X.columns.tolist(), dtype=object)
elif not hasattr(self, "feature_names_in_"):
     ...
```

`validate_data` has already rebound `X` to a NumPy array by then, so
`hasattr(X, "columns")` is always `False` and the first branch is dead. It reads
as though it handles DataFrame input; it does not, and it is not what makes
DataFrame input work (scikit-learn's own `validate_data` sets `feature_names_in_`
before this code runs).

### Where

| File | Locate with |
|---|---|
| `philanthropy/preprocessing/_wealth_percentile.py` | `grep -n "we should probably warn or raise" philanthropy/preprocessing/_wealth_percentile.py` |
| `philanthropy/preprocessing/_wealth_percentile.py` | `grep -n 'hasattr(X, "columns")' philanthropy/preprocessing/_wealth_percentile.py` |
| `tests/test_preprocessing.py` | `grep -n "WealthPercentileTransformer" tests/test_preprocessing.py` |

### What to change

1. **Decide warn or raise, and justify it on the issue before you code.** The
   house style is worth checking first: `GratefulPatientFeaturizer` warns and
   returns zeros when it cannot find its merge key (`grep -n "warnings.warn"
   philanthropy/preprocessing/_grateful_patient.py`), while structurally invalid
[...]

PR #175, opened 2026-09-06
fix: remove dead hasattr branch in WealthPercentileTransformer.fit (#168)

### LINKED ISSUE #168: cleanup: `WealthPercentileTransformer.fit` has a branch that can never run
### What's wrong

deleted.

```
```

### Where

| File | Locate with |
|---|---|

### What to change

### Tests to add or extend

### Done when

```bash

python -c "
"
make ci
make riskcov
```

### First time here?

HeaTTap:
## What & why

Resolves #168

In WealthPercentileTransformer.fit, validate_data returns a NumPy array so hasattr(X, "columns") was never True and that branch never executed. The elif branch beneath it was handling both DataFrame and ndarray inputs.

This PR:
1. Deletes the dead hasattr(X, "columns") branch in WealthPercentileTransformer.fit and promotes the elif condition to a regular if statement with an explanatory comment.
2. Adds unit test test_wealth_percentile_feature_names_match_for_frame_and_array_input in tests/test_preprocessing.py verifying that get_feature_names_out preserves column names for DataFrame inputs and generates x0..xn for ndarray inputs.
3. Adds entry to CHANGELOG.md and CONTRIBUTORS.md.

## Checklist
- [x] `make ci` passes locally (lint -> collection -> tests -> coverage >= 92%)
- [x] New/changed public API has docstrings and is exported in the subpackage `__init__.py`
- [x] Tests added or updated
- [x] `CHANGELOG.md` updated under `[Unreleased]`
- [x] Added yourself to `CONTRIBUTORS.md` (skip if you would rather not be listed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
+### Fixed
+- Removed dead `hasattr(X, "columns")` branch in `WealthPercentileTransformer.fit` since `validate_data` returns a NumPy array and sets `feature_names_in_` for DataFrame inputs (issue #168).
+
 ### Changed
 - README coverage badge now links to `pyproject.toml` and reads "≥92% floor"
   rather than a bare "≥92%". It was a static shields.io string with no tie to
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -45,6 +45,8 @@ contribution. Code, docs, tests, and review all count.
   single-class fallback coverage for
   `PlannedGivingIntentScorer.predict_intent_score`
   ([#149](https://github.com/PhilanthroPy-Project/PhilanthroPy/pull/149)).
+- [@HeaTTap](https://github.com/HeaTTap): removed dead `hasattr(X, "columns")` branch in `WealthPercentileTransformer.fit` and pinned feature-name handling for DataFrame and array inputs ([#168](https://github.com/PhilanthroPy-Project/PhilanthroPy/issues/168)).
+
 
 ## Getting listed
 
--- a/philanthropy/preprocessing/_wealth_percentile.py
+++ b/philanthropy/preprocessing/_wealth_percentile.py
@@ -42,11 +42,9 @@ class WealthPercentileTransformer(TransformerMixin, BaseEstimator):
         training distribution, not against the batch being transformed.
         """
         X = validate_data(self, X, ensure_all_finite="allow-nan", reset=True)
-        
-        if hasattr(X, "columns"):
-             self.feature_names_in_ = np.array(X.columns.tolist(), dtype=object)
-        elif not hasattr(self, "feature_names_in_"):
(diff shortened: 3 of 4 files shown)


shivamlalakiya:
Thanks for this, the fix matches #168 exactly. There's a merge conflict now against `main` (PR #174, which pins the same class's edge-case tests, merged first), so this needs a rebase before it can go in. Both PRs happened to touch the same three spots, and the fix in each case is to keep both sides, not pick one:

**`tests/test_preprocessing.py`**
- The `WealthPercentileTransformer` import ends up listed twice — delete one.
- Both PRs add a test right after `test_crm_cleaner_on_unnamed_ndarrays_skips_named_columns`: #174 adds a `class TestWealthPercentileTransformer` with two tests, this PR adds a standalone `test_wealth_percentile_feature_names_match_for_frame_and_array_input`. Keep both; I'd suggest moving your standalone test into that class (as a third method, dropping the `wealth_percentile_` prefix since it's now scoped) so all three live together, but a standalone function next to the class works too.

**`CONTRIBUTORS.md`**
- Keep both new lines (yours and Lars's from #169); no actual conflict in content, just adjacent insertions.

**`CHANGELOG.md`**
- Same as above, auto-merges cleanly, just confirm both entries are present after rebase.

```bash
git fetch origin
git rebase origin/main
# resolve the three spots above, then:
git add -A && git rebase --continue
git push --force-with-lease
```

Once that's green I'll merge it. Nice first PR, thanks for tightening this up.

shivamlalakiya:
Thank you for this, @HeaTTap, and sorry for the outcome: the fix is correct but it has landed by another route.

Issue #168 is now closed by #179, which was merged as 6dfafa5. That PR removed the same dead `hasattr(X, "columns")` branch as part of a larger change to `WealthPercentileTransformer.fit` for #156, so the one-line deletion here is already on `main`. Your diagnosis was right, and the `get_feature_names_out` assertion you added is equivalent to the coverage that #174 and #179 landed.

Two contributors independently opened a fix for #168 the same evening (#175 and #176), which is our process failure, not yours. We are adding a "claim the issue before you open a PR" rule to CONTRIBUTING.md so this stops happening.

You are owed a CONTRIBUTORS.md line and do not have one yet. That is tracked and will be added in a follow-up PR, not dropped.

Closing as subsumed by #179.

shivamlalakiya:
@HeaTTap following up on this one properly rather than leaving it as a silent close.

You and another contributor solved the same four-line problem on the same evening, and closing yours cost you that evening for nothing. That was the project's process failure, not yours. There was no way to claim an issue before opening a pull request, so there was no way for either of you to know the other was working on it. The only person who could see both was me, and I didn't act on it fast enough.

That's fixed now. #187 is merged and adds a claim-before-PR rule to CONTRIBUTING.md, so an issue gets assigned before anyone writes code, and it credits you by name in CHANGELOG.md for the fix attempt.

If you want another one, the `good first issue` label has open ones and I'll assign whichever you comment on within 24 hours. And if the setup or the contribution process was annoying in any other way, I'd genuinely like to hear it: https://github.com/PhilanthroPy-Project/PhilanthroPy/discussions/188

PR #176, opened 2026-09-06
fix: remove dead branch in WealthPercentileTransformer.fit (#168)

### LINKED ISSUE #168: cleanup: `WealthPercentileTransformer.fit` has a branch that can never run

Larslllllll:
## Summary

The `hasattr(X, "columns")` branch in `fit()` never executed because `validate_data()` returns a NumPy array, which never has a `.columns` attribute. `validate_data()` itself sets `feature_names_in_` when given a DataFrame, so the column names are preserved regardless.

The `elif` branch (array-input path) was doing all the work. This PR deletes the dead branch and promotes the `elif` to a plain `if` with an explanatory comment.

## Changes

- **Deleted** the dead `if hasattr(X, "columns"):` branch
- **Promoted** the `elif` to a plain `if` with comment explaining why the check is on `self` not `X`
- **Added** test `test_wealth_percentile_feature_names_match_for_frame_and_array_input` to pin expected feature names for both DataFrame and array inputs

## Verification

```bash
# Verify dead branch is gone

# Verify behavior is preserved
python -c "
assert list(a.get_feature_names_out()) == ['net_worth', 'other', 'net_worth_pct_rank']
assert list(b.get_feature_names_out()) == ['x0', 'x1']
"
```

Fixes #168

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   project's standing rules.
 
 ### Added
+- Tests pinning the two untested branches of `WealthPercentileTransformer`:
+  the all-missing column path (returns NaN ranks, keeps a stable output
+  width, raises no warning) and the partially-missing column path (NaN
+  input rows get NaN rank, observed rows get numeric rank). Closes #169.
 - **`datasets.make_donor_panel`** (Tier 2, Beta): a seeded multi-year donor
   panel returning gift-level rows rather than one aggregated row per donor.
   `generate_synthetic_donor_data` cannot demonstrate `RFMTransformer` (needs a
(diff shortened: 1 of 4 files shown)


shivamlalakiya:
Thank you, @Larslllllll. Closing this one as already-landed rather than rejected, and your credit is safe on both counts.

Where each piece went:
- The edge-case tests here (all-missing and partially-missing wealth columns) were merged as #174 and are on `main` now, along with your CONTRIBUTORS.md entry.
- The dead `hasattr(X, "columns")` branch removal was merged as part of #179 (6dfafa5), which closed #168 alongside #156.

So everything in this PR is on `main`, which is also why it now shows as conflicting: it is trying to re-add lines that are already there.

The duplication with #175 the same evening was our process failure, not yours. A "claim the issue before you open a PR" rule is going into CONTRIBUTING.md so contributors stop colliding like this.

Closing as already merged via #174 and #179.

Issue #110, opened 2026-08-21
GratefulPatientFeaturizer/EncounterTransformer: index-encounter keying vs. the PHI-free bundle contract need a design decision


shivamlalakiya:
### The conflict

`GratefulPatientFeaturizer` and `EncounterTransformer` key on the donor's
**latest** discharge (`grep -n "merge_key" philanthropy/preprocessing/_grateful_patient.py`,
groupby on `merge_key` then take `last_discharge`). For a donor with more than
one encounter, a gift is measured against whichever discharge happens to be
most recent at fit/transform time, not the discharge that was actually closest
in time to that specific gift. Recurrent patients (oncology, transplant) are
the people this most misrepresents.

The fix needs the raw per-gift, per-encounter rows available at `transform`
time so each gift can be joined to its own nearest prior discharge. That is
exactly what `__getstate__` is designed to strip
(`grep -n "__getstate__" -A8 philanthropy/preprocessing/_grateful_patient.py`):
raw `encounter_df` is PHI-bearing and is replaced with `None` on serialisation
so a saved model bundle never carries patient rows (#78).

Per-gift keying and the PHI-free bundle contract are in direct conflict. This
is not a bug with an obvious patch; it needs a design decision on which
guarantee gives way, or a third option (e.g. storing only a de-identified
per-donor encounter-to-gift offset at fit time, if that is precise enough).

### Why it matters now

This is the one item from the 2026-08-13 audit that landed as "documented
limitation" rather than "fixed" going into the JOSS submission. It affects the
grateful-patient timing story the paper's Software design section describes as
solved via `as_of`; per-gift keying is a separate, still-open gap in the same
story.

### What would close this

A written decision (this issue, a discussion, or a design doc) on one of:

1. Keep latest-discharge keying, document the recurrent-patient
   misattribution explicitly as an accepted limitation in
   `docs/explanation/compliance_considerations.md`.
2. Store a de-identified per-donor "days since nearest prior discharge, as of
   each gift date" summary at `fit` time (no raw dates, no diagnosis), so
   `__getstate__` still has nothing PHI-bearing to strip.
3. Some other resolution.

This needs a person to decide the tradeoff; it cannot be closed by a PR alone.

shivamlalakiya:
Proposal, since the issue asks for a decision even if it's a documented deferral.

The real constraint: per-gift nearest-prior-discharge keying needs the raw encounter
[...]

PR #180, opened 2026-09-06
fix: accept parsed gift dates

### LINKED ISSUE #163: EncounterTransformer rejects a parsed datetime64 gift_date and accepts only date strings
### What's wrong

`EncounterTransformer.fit_transform` raises on a gift frame whose `gift_date` column is an actual
`datetime64` column, and succeeds only when the dates are **strings**. For a library whose central
argument is about temporal correctness, rejecting parsed dates and accepting unparsed ones is
backwards: `pd.read_csv(..., parse_dates=["gift_date"])`, any SQL loader, and this package's own
`philanthropy.datasets.make_donor_panel` all hand you `datetime64`.

```python
import pandas as pd
from philanthropy.preprocessing import EncounterTransformer

enc = pd.DataFrame({"donor_id": [1, 2], "discharge_date": ["2023-05-10", "2022-11-01"]})
gifts = pd.DataFrame({
    "donor_id":    [1, 2],
    "gift_date":   pd.to_datetime(["2023-08-15", "2023-03-20"]),   # parsed, as any real loader gives
    "gift_amount": [5000.0, 500.0],
})

EncounterTransformer(
    encounter_df=enc, discharge_col="discharge_date",
    gift_date_col="gift_date", merge_key="donor_id",
).fit_transform(gifts)
```

```
numpy.exceptions.DTypePromotionError: The DType <class 'numpy.dtypes.Float64DType'> could not be
promoted by <class 'numpy.dtypes.DateTime64DType'>. ... The full list of DTypes is:
(<class 'numpy.dtypes.Int64DType'>, <class 'numpy.dtypes.DateTime64DType'>,
 <class 'numpy.dtypes.Float64DType'>)
```

Change one line, `gift_date=gifts["gift_date"].dt.strftime("%Y-%m-%d")`, and the same call returns a
`(2, 3)` frame. The transformer's own docstring example passes strings, which is why this has never
been noticed: with `int` + `str` + `float` the frame collapses to `object` dtype and numpy has
nothing to promote.

### Why it matters

The error message names numpy dtypes and points at nothing the caller wrote. A user whose CSV loader
happens to parse dates gets a traceback from `sklearn.utils.validation` with no indication that the
fix is to *un*-parse a column, which is not advice anyone would guess or want to follow.

It also blocks the obvious composition. `make_donor_panel(include_encounters=True)` produces exactly
the two frames this transformer is for, and they cannot be fed to it without a `strftime` round-trip
first. `tests/test_donor_panel.py::test_encounters_feed_the_encounter_transformer_with_an_as_of_cutoff`
currently carries a comment explaining the workaround; that comment should be deletable.

### Where

| What | Locate with |
|---|---|
| the validation call that promotes dtypes | `grep -n 'validate_data\|check_array' philanthropy/preprocessing/_encounters.py` |
| the docstring example that passes strings | `grep -n 'gift_date' philanthropy/preprocessing/_encounters.py \| head` |
| the sibling that may have the same problem | `philanthropy/preprocessing/_encounter_recency.py`, `_discharge_window.py` |
| the test carrying the workaround | `grep -n 'strftime' tests/test_donor_panel.py` |

Check the two siblings before writing a fix. If they share the pattern, fix it once where all three
route through rather than three times; if they differ, the difference is worth a comment.

### What to change

The frame is mixed-dtype by nature (an identifier, a date, an amount), so it should be validated as
`dtype=object` or the date column should be separated before validation rather than pushed through a
numeric `check_array`. Whichever route, the fix should:

1. accept `datetime64` `gift_date` **and** keep accepting strings, since the docstring, the doctests
   and existing users all pass strings;
2. keep the `as_of` cutoff comparison working on both;
3. raise something that names the column and the expected type if a genuinely unusable dtype arrives,
   instead of surfacing numpy's promotion error.

### Tests to add

- Parametrise an existing `EncounterTransformer` test over `["string", "datetime64"]` gift dates and
  assert the two produce **equal** output, not merely that both run.
- The same for whichever of `EncounterRecencyTransformer` /
  `DischargeToSolicitationWindowTransformer` turns out to share the bug.
- A negative test that an unusable dtype raises a message naming the column.

### Done when

```bash
# Exits 0 only once parsed dates work.
python -c "
import pandas as pd
from philanthropy.preprocessing import EncounterTransformer
enc = pd.DataFrame({'donor_id': [1, 2], 'discharge_date': ['2023-05-10', '2022-11-01']})
g = pd.DataFrame({'donor_id': [1, 2],
                  'gift_date': pd.to_datetime(['2023-08-15', '2023-03-20']),
                  'gift_amount': [5000.0, 500.0]})
t = EncounterTransformer(encounter_df=enc, discharge_col='discharge_date',
                         gift_date_col='gift_date', merge_key='donor_id')
t.set_output(transform='pandas')
out = t.fit_transform(g)
assert 'days_since_last_discharge' in out.columns, out.columns
"
make ci
```

and the `strftime` workaround comment in `tests/test_donor_panel.py` is gone.

### Found by

Writing `make_donor_panel` (#162). The generator produces the encounter and gift tables this
transformer exists for, and they could not be handed to it directly.

be-student:
Fixes #163.

Normalizes the gift-date column before scikit-learn validates the mixed-dtype frame, so both strings and `datetime64` inputs work. Invalid non-missing dates now raise a column-specific error.

Validation: `make ci` (1,951 passed, 25 skipped; 97.44% coverage).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,10 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   wheel and turn this into a second working-tree test. Windows is included
   because the main matrix is Linux plus macOS.
 ### Fixed
+- `EncounterTransformer` now accepts parsed `datetime64` gift dates alongside
+  date strings, and reports unparseable values with the configured gift-date
+  column name instead of exposing NumPy's mixed-dtype promotion error. Closes
+  #163.
 - `CRMCleaner.transform` no longer silently corrupts complex amounts into
   wrong finite floats: cells holding actual `complex` values are masked to
   NaN with a `UserWarning` naming them, and a column where nothing parses
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -49,6 +49,9 @@ contribution. Code, docs, tests, and review all count.
   unit-test coverage for `WealthPercentileTransformer`'s all-missing and
   partially-missing column branches in `WealthPercentileTransformer` (closes
   [#169](https://github.com/PhilanthroPy-Project/PhilanthroPy/issues/169)).
+- [@be-student](https://github.com/be-student): made `EncounterTransformer`
+  accept parsed gift dates and report invalid dates with a column-specific
+  error (closes [#163](https://github.com/PhilanthroPy-Project/PhilanthroPy/issues/163)).
 
 ## Getting listed
 
--- a/philanthropy/preprocessing/_encounters.py
+++ b/philanthropy/preprocessing/_encounters.py
@@ -338,6 +338,18 @@ class EncounterTransformer(TransformerMixin, BaseEstimator):
                     f"the `{label}` parameter in EncounterTransformer."
                 )
 
+    def _parse_gift_dates(self, X: pd.DataFrame) -> pd.Series:
+        """Return parsed gift dates or raise a column-specific error."""
+        values = X[self.gift_date_col]
+        parsed = pd.to_datetime(values, errors="coerce")
+        invalid = values.notna() & parsed.isna()
+        if invalid.any():
+            raise ValueError(
+                f"Column {self.gift_date_col!r} must contain date-like values; "
+                f"could not parse {int(invalid.sum())} non-missing value(s)."
+            )
+        return parsed
+
 
     # ------------------------------------------------------------------
     # Column-drop utilities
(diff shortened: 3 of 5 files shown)


shivamlalakiya:
@be-student three merged fixes in one day, and all three were real defects rather than coverage padding. #179 in particular closed a silent no-op: a `wealth_cols` list that matched no column was accepted and quietly ignored, so a user would have got numbers back with no indication that the transformer had done nothing. That's the failure mode I most want caught, because nothing downstream ever raises.

You clearly read the codebase rather than just the issue text, so I'd rather point you at something open-ended than another labelled issue.

#110 is the one unresolved design question in the hospital-fundraising path: index-encounter keying versus the PHI-free bundle contract, which `tests/test_no_phi_in_bundles.py` currently depends on. I've written up why it's deferred rather than decided, but a deferral I argued with myself about is worth less than one someone else has pushed back on. If you have a view, the thread is the right place for it.

Separately, if anything about getting set up here was annoying, I'd like to hear that too: https://github.com/PhilanthroPy-Project/PhilanthroPy/discussions/188

PR #186, opened 2026-09-06
fix: WealthPercentileTransformer silent no-op and dead branch, EncounterTransformer datetime64 rejection

### LINKED ISSUE #156: `WealthPercentileTransformer` silently ignores a `wealth_cols` list that matches no column

### LINKED ISSUE #163: EncounterTransformer rejects a parsed datetime64 gift_date and accepts only date strings

### LINKED ISSUE #168: cleanup: `WealthPercentileTransformer.fit` has a branch that can never run

shivamlalakiya:
## Summary
Three real defects found during a pass over the JOSS-prep plan, plus the process fix for how the third one happened.

- `WealthPercentileTransformer(wealth_cols=[...])` now raises `ValueError` instead of silently training without the wealth signal when none of the named columns exist in `X`. A typo'd column name was indistinguishable from an intentional no-op. Closes #156.
- `WealthPercentileTransformer.fit`'s dead `hasattr(X, "columns")` branch is removed (`validate_data` already returns a NumPy array and already sets `feature_names_in_` for DataFrame input). Two contributors independently fixed this same branch the same evening, #175 (@HeaTTap) and #176 (@Larslllllll); this PR lands that fix and credits both in `CONTRIBUTORS.md`. Closes #168.
- `EncounterTransformer` no longer rejects a `gift_date` column already parsed to `datetime64`. Any real loader (`pd.read_csv(parse_dates=...)`, a SQL read, this package's own `make_donor_panel`) hands back exactly that column mix, and `validate_data`'s numpy promotion of `datetime64` against `float64` raised before the transformer got a chance to read it. Closes #163.
- `CONTRIBUTING.md` gets a "Claiming an issue" rule (comment before you open a PR, wait for assignment) so the #175/#176 duplication doesn't repeat; the issue-draft template states it inline too.

Also filed 5 good-first-issue drafts from `scripts/issue-drafts/` as real GitHub issues (#181-#185), each re-verified against current line numbers/greps first. Not part of this diff since they're process, not code.

## Test plan
- [x] New regression tests for all three fixes: `test_unmatched_wealth_cols_raises`, `test_feature_names_match_for_frame_and_array_input`, `test_accepts_parsed_datetime64_gift_date`
- [x] `make ci` green
- [x] `make riskcov` green (97% branch over the risk-tier subtree, 93% floor)
- [x] Full suite green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,30 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
+### Fixed
+- `WealthPercentileTransformer.fit`'s dead `hasattr(X, "columns")` branch is
+  removed: `validate_data` already returns a NumPy array and already sets
+  `feature_names_in_` for DataFrame input, so the branch never ran. Reported
+  as #168 and independently fixed by two duplicate PRs on the same evening,
+  #175 (@HeaTTap) and #176 (@Larslllllll); this change lands that fix and
+  credits both. See "Claiming an issue" in `CONTRIBUTING.md` for the process
+  fix that should prevent the next one.
+- `WealthPercentileTransformer(wealth_cols=[...])` now raises `ValueError`
+  instead of silently doing nothing when none of the named columns exist in
+  the training frame. A caller who names specific columns and gets none of
+  them is one typo away from training a model silently missing its wealth
+  signal. Closes #156.
+- `EncounterTransformer` no longer rejects a `gift_date` column already
+  parsed to `datetime64`, which previously raised
+  `numpy.exceptions.DTypePromotionError` deep inside `validate_data` (numpy
+  cannot promote `datetime64` and `float64` into one array dtype). A
+  DataFrame mixing a datetime64 column with numeric ones is now cast to
+  `object` dtype first. Any real loader
+  (`pd.read_csv(..., parse_dates=[...])`, a SQL read, this package's own
+  `make_donor_panel`) hands back exactly this column mix, so the previous
+  behavior rejected the common case and accepted only pre-formatted date
+  strings. Closes #163.
+
 ### Changed
 - README coverage badge now links to `pyproject.toml` and reads "≥92% floor"
   rather than a bare "≥92%". It was a static shields.io string with no tie to
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,6 +11,14 @@ done. A first PR does not need to be big; docs fixes and missing tests are
 genuinely wanted. Comment on the issue to claim it, and ask there if anything in
 this guide does not work; a question is a valid contribution too.
 
+### Claiming an issue
+
+Comment on the issue before you open a pull request, and wait for a maintainer
+to assign it to you; we assign within 24 hours. This is not bureaucracy:
+issues #175 and #176 were two people solving the same four-line problem on the
+same evening, and one of them wasted their time. If an issue already has an
+assignee, pick another one; the `good first issue` label always has open ones.
+
 ## Setup
 
 Fork the repository on GitHub first; you will not have push access to
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -48,7 +48,18 @@ contribution. Code, docs, tests, and review all count.
 - **Lars** ([@Larslllllll](https://github.com/Larslllllll)): added the missing
   unit-test coverage for `WealthPercentileTransformer`'s all-missing and
   partially-missing column branches in `WealthPercentileTransformer` (closes
-  [#169](https://github.com/PhilanthroPy-Project/PhilanthroPy/issues/169)).
+  [#169](https://github.com/PhilanthroPy-Project/PhilanthroPy/issues/169)), and
+  identified the dead `hasattr(X, "columns")` branch in
+  `WealthPercentileTransformer.fit`
+  ([#176](https://github.com/PhilanthroPy-Project/PhilanthroPy/pull/176), closes
+  [#168](https://github.com/PhilanthroPy-Project/PhilanthroPy/issues/168)).
+- [@HeaTTap](https://github.com/HeaTTap): independently identified and fixed the
+  same dead `hasattr(X, "columns")` branch in `WealthPercentileTransformer.fit`
+  ([#175](https://github.com/PhilanthroPy-Project/PhilanthroPy/pull/175), closes
+  [#168](https://github.com/PhilanthroPy-Project/PhilanthroPy/issues/168)). Two
+  people fixing the same four-line problem on the same evening was the
+  project's process failure, not either contributor's; see "Claiming an issue"
+  in `CONTRIBUTING.md`.
 
 ## Getting listed
 
--- a/philanthropy/preprocessing/_encounters.py
+++ b/philanthropy/preprocessing/_encounters.py
@@ -91,6 +91,22 @@ def _apply_as_of_cutoff(
     return enc[keep]
 
 
+def _avoid_dtype_promotion(X: Any) -> Any:
+    """Cast a DataFrame with a parsed datetime column to ``object`` dtype.
+
+    ``validate_data`` converts ``X`` to a single-dtype array, and numpy cannot
+    promote ``datetime64`` together with ``int``/``float`` into one dtype. A
+    caller's own loader (``pd.read_csv(parse_dates=...)``, a SQL read,
+    ``philanthropy.datasets.make_donor_panel``) hands back exactly this
+    column mix, so ``gift_date_col`` must not need pre-formatting to strings.
+    """
+    if isinstance(X, pd.DataFrame) and any(
+        pd.api.types.is_datetime64_any_dtype(dt) for dt in X.dtypes
+    ):
+        return X.astype(object)
+    return X
+
+
 def _warn_if_unbounded(
     enc: pd.DataFrame, discharge_col: str, gift_dates: Any, class_name: str
 ) -> None:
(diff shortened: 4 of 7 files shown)


shivamlalakiya:
Closing this in favor of already-open external contributor PRs for the same issues: #180 (@be-student, issue #163) and #179 (@be-student, issue #156, which also removes the #168 dead branch). Those landed first, are already green, and external contribution signal is a tracked JOSS blocker, so they should merge over a duplicate from me.

The parts of this PR not covered by #179/#180 (the CONTRIBUTING.md "Claiming an issue" rule and the issue-draft template line) are being handled separately to avoid a third duplicate.

Issue #22, opened 2026-08-02
docs: two commands on the "Develop and Test" page point at test files that do not exist


shivamlalakiya:
**Type:** documentation · **Good first issue** · **No domain knowledge needed** · **~30 minutes including setup**

## What's wrong

`docs/how-to/develop_and_test.md` is the only development page on the published docs site (it's in the nav at `mkdocs.yml:89`). It contains four `bash` commands. Two of them name test files that don't exist, and one silently lowers the coverage gate.

Nothing catches this today: `tests/test_doc_examples.py` executes the fenced **`python`** blocks in every docs page, but this page has only **`bash`** blocks, so it is never collected (see the filter at `tests/test_doc_examples.py:40`).

A contributor who follows this page instead of `CONTRIBUTING.md` runs two commands that collect zero tests, verifies coverage against a threshold 7 points below the real gate, and never learns that `make ci` is the actual gate.

## The four edits

All in one file: `docs/how-to/develop_and_test.md`.

### 1. Line 30 — file does not exist

```bash
pytest tests/test_sklearn_compat.py -v
```

`tests/test_sklearn_compat.py` is not in the repo (there is a stale `.pyc` for it under `tests/__pycache__/`, which is why it may look like it exists — that directory is gitignored build junk). The real file is `tests/test_sklearn_compliance.py`.

Change the filename to:

```bash
pytest tests/test_sklearn_compliance.py -v
```

*(Verified: `963 passed` in ~20s.)*

### 2. Line 38 — file does not exist

```bash
pytest tests/test_transformers_property.py -v --hypothesis-seed=12345
```

`tests/test_transformers_property.py` is not in the repo either. The property-based tests now live in `tests/test_properties.py`, which describes itself as "the single home for property-based (Hypothesis) tests over the transformers."

Change the filename to:

```bash
pytest tests/test_properties.py -v --hypothesis-seed=12345
```

*(Verified: `18 passed` in ~15s with that seed.)*

> Out of scope: `tests/test_preprocessing_properties.py` also still exists even though `test_properties.py`'s docstring says it was absorbed. That's my mess to sort out, not yours — please don't touch it in this PR.

### 3. Lines 19 and 22 — stale coverage threshold

Line 19 says "confirm coverage stays above 85%" and line 22 passes `--cov-fail-under=85`.

[...]

Issue #21, opened 2026-08-02
docs: PR checklist and AGENTS.md still say the coverage gate is 85%; it is 92%


shivamlalakiya:
## The problem

The coverage gate is **92%** overall, with a **93%** floor on the risk tier. Three docs files still say **85%**.

Source of truth:

- `pyproject.toml:113` → `fail_under = 92`
- `Makefile:6` → `RISK_FLOOR := 93`, run by `make riskcov` (`.github/workflows/ci.yml:76`)

The most damaging copy is the PR template. It is the last thing you read before hitting "Create pull request", and it asks you to tick a box certifying `coverage ≥ 85%` — which you can honestly tick at 88%, and then CI fails you on a gate you were never shown.

This is a docs-only change. **Three numbers, two files.** No Python is touched.

### 1. `.github/PULL_REQUEST_TEMPLATE.md` line 6

```
- [ ] `make ci` passes locally (lint → collection → tests → coverage ≥ 85%)
```

Change `85%` → `92%`.

### 2. `AGENTS.md` lines 47-48

```
4. Write the tests. 5. Run `make ci` (collection → full suite → coverage ≥ 85%).
Never `git push --no-verify`; the coverage gate is 85% and must stay green.
```

Change **both** `85%`/`85` → `92%`/`92`.

## Please do NOT touch these

- **`docs/how-to/develop_and_test.md`** — it carries the same stale `85`, but it is claimed by a sibling issue that rewrites four commands on that page at once. Leave the file alone so the two PRs do not collide.
- **`CHANGELOG.md:380`** — `Coverage gate: pytest --cov-fail-under=85` sits under `## [0.2.0] - 2026-03-14`. That is a historical record of what 0.2.0 actually shipped. It was 85% at the time and it stays 85%.
- **`README.md:13`** — the badge already reads `≥92%`. Correct, leave it.
- **`CONTRIBUTING.md`** — already states the 92% gate correctly. Leave it.

## Acceptance

From the repo root:

```bash
# 1. No stale 85 in either file. Prints nothing, exits 1.
grep -rn '85' .github/PULL_REQUEST_TEMPLATE.md AGENTS.md

# 2. The 85s became 92s rather than vanishing. Must total 3:
#    PULL_REQUEST_TEMPLATE.md:1, AGENTS.md:2
grep -rc '92' .github/PULL_REQUEST_TEMPLATE.md AGENTS.md

# 3. Sanity-check the source of truth you are matching
[...]

PR #215, opened 2026-09-10
Document Windows-specific gaps in the local test/CI gate

### LINKED ISSUE #195: docs: the local gate is Unix-only; `make ci`, `make riskcov` and `sh scripts/install_hooks.sh` have no documented Windows path
### What's wrong

Every setup instruction in this repository assumes a Unix shell with `make` on the PATH. Windows ships neither, so a Windows contributor hits two dead ends before their first green run and has to reverse-engineer the gate from the `Makefile`.

Reported by @AsavariCharati on Discussion #188 and on #74, after already landing #74:

1. `make ci` and `make riskcov` do not run, because `make` is not installed on Windows by default. They ran each Makefile command by hand, one at a time, working out the order themselves.
2. `sh scripts/install_hooks.sh` does not run in PowerShell. It needs Git Bash, which nothing in `CONTRIBUTING.md` says.

Neither blocked them, which is the point: this is invisible friction that costs a first-time contributor an hour and never shows up in CI, because CI is Ubuntu and macOS only.

Docs only. No behaviour change, and deliberately **not** a Windows `Makefile` or a `make.bat`: a second copy of the gate is a second thing to drift, and #21 and #22 exist because this project has already been bitten by exactly that.

### Where

| File | Locate with |
|---|---|
| `CONTRIBUTING.md` | `grep -n "install_hooks\|make ci\|make riskcov" CONTRIBUTING.md` |
| `docs/how-to/develop_and_test.md` | `grep -n "make ci\|pytest" docs/how-to/develop_and_test.md` |
| `Makefile` (read only, do not copy its recipes into the docs) | `grep -n "^ci:\|^riskcov:\|RISK_TIER\|RISK_FLOOR" Makefile` |

### What to change

1. In `CONTRIBUTING.md`, next to the `sh scripts/install_hooks.sh` line in the clone block, say that this step needs a Unix shell and that Windows users should run it from Git Bash rather than PowerShell. Say the same thing at the second mention further down the file (the `grep` above finds both).
2. While you are there, check and state whether the *installed* hook then fires from PowerShell. Git for Windows executes `#!/bin/sh` hooks through its own bundled bash, so it should, but confirm it on your machine before writing it down. If it does, that sentence is the whole answer to friction 2: one step needs Git Bash, not the whole workflow.
3. In the "Before pushing any commit" section of `CONTRIBUTING.md`, add a short note for machines without `make`: the six numbered steps already listed there map one-to-one onto targets in the `Makefile` (`lint`, `typecheck`, `check`, `doctest`, `coverage`, then `riskcov`), and each target's recipe can be read and run directly out of that file.
4. **Do not paste the recipes into the docs.** In particular do not copy the `RISK_TIER` include list or the `RISK_FLOOR` number: the `Makefile` is the single source of truth for both, CI runs the same target, and a second copy in prose is how the 85% versus 92% drift in #21 happened. Point at the file, do not duplicate it.
5. In `docs/how-to/develop_and_test.md`, add the one gate that page is missing. It documents the coverage, compliance, property and doctest runs, but not the risk-tier floor, which is the check most likely to turn a first PR red. One sentence plus a pointer to `make riskcov` and its `Makefile` recipe.

### Tests to add or extend

None. This change touches no code under `philanthropy/`, so the credit guard's changelog requirement does not fire on it, but `AGENTS.md` still asks every PR for a `## [Unreleased]` entry in `CHANGELOG.md` and a line in `CONTRIBUTORS.md`. Add both.

### Done when

```bash
# Exits 0 only once both files name the make-free path. Run before opening the PR.
grep -qi 'git bash' CONTRIBUTING.md
grep -q 'riskcov' docs/how-to/develop_and_test.md
# And the number stayed in one place: the floor is still only in the Makefile.
test "$(grep -rc 'fail-under=93\|RISK_FLOOR := 93' Makefile)" -ge 1
test "$(grep -rl '93' CONTRIBUTING.md docs/how-to/develop_and_test.md | wc -l)" -eq 0
```

Comment here to claim it before opening a PR (see "Claiming an issue" in `CONTRIBUTING.md`). If you are on Windows, you are the right person to write this: say what actually worked on your machine rather than what should work on it.

The third item @AsavariCharati reported, `pytest --cov` raising `ImportError: cannot load module more than once per process` on the numpy import while `coverage run -m pytest` gave identical numbers, is a separate and more interesting problem and is not in scope here.

AsavariCharati:
Resolves #195

Documents the Windows-specific gaps in the local test/CI workflow, reported firsthand across #74 and #188:

- `CONTRIBUTING.md`: adds a callout after Setup noting `make`/`sh` aren't available in PowerShell by default, plus a new "Testing on Windows" section with the equivalent `python -m` commands for `make ci`, and a pointer to the Makefile for the risk-tier coverage floor (not copied verbatim, per the issue's instruction).
- `docs/how-to/develop_and_test.md`: adds a short Windows note after Prerequisites covering the `pytest-cov` / numpy `ImportError` workaround.

Also confirmed by testing (not assumed): the installed pre-push hook does **not** fire when pushing from plain PowerShell -only from a POSIX-compatible shell (verified with a throwaway branch/commit/push, since removed).

No code changes, so the CHANGELOG credit-guard doesn't fire on this PR, but added an entry under [Unreleased] → Changed and a line in CONTRIBUTORS.md anyway, per the issue's explicit instruction.

Note: while running the full suite locally to verify nothing broke, found 11 pre-existing failures unrelated to this change - `urlparse` on Windows reads a `C:\...` path's drive letter as a URL scheme (`'c'`), so `ensure_local_path` rejects every valid local Windows path. Out of scope here; happy to open a separate issue for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,13 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   lifetime-giving-greater-than-zero variable, where the feature was the
   outcome. It replaces the synthetic `total_lifetime_giving` sketch and gains a
   runnable `as_of` walkthrough of the same failure. Closes #209.
-
+- `CONTRIBUTING.md` and `docs/how-to/develop_and_test.md` now document the
+  Windows-specific gaps in the local test/CI gate: `make ci`/`make riskcov`
+  require `make`, which is not available in PowerShell by default, and
+  `sh scripts/install_hooks.sh` requires Git Bash. Testing confirmed the
+  installed pre-push hook does not fire when pushing from plain PowerShell.
+  Closes #195.
+  
 ### Fixed
 - `EncounterRecencyTransformer` no longer catches timezone conversion errors
   and retries with `tz_localize` after parsing with `utc=True`. The retry was
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,7 +54,7 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   `sh scripts/install_hooks.sh` requires Git Bash. Testing confirmed the
   installed pre-push hook does not fire when pushing from plain PowerShell.
   Closes #195.
-  
+
 ### Fixed
 - `EncounterRecencyTransformer` no longer catches timezone conversion errors
   and retries with `tz_localize` after parsing with `utc=True`. The retry was
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -131,9 +131,15 @@ equivalent commands directly instead of `make ci` / `make riskcov`:
 python -m flake8 philanthropy tests examples
 python -m mypy philanthropy
 python -m pytest philanthropy --doctest-modules -q --no-cov
+python -m pytest tests/ --collect-only -q
 python -m coverage run -m pytest tests/ -q
-python -m coverage report --fail-under=92
+python -m coverage report
 ```
+
+`coverage report` reads the overall floor from `[tool.coverage.report]` in
+`pyproject.toml`, so the bare command enforces exactly what CI does. Do not
+pass `--fail-under` here: the number is defined once and copying it into a
+second place is how the two drift apart.
 
 For the risk-tier coverage floor, see the `RISK_TIER`/`RISK_FLOOR` values
 in the `Makefile` and run:
@@ -144,8 +150,8 @@ python -m coverage report --include='<RISK_TIER from Makefile>' --fail-under=<RI
 
 If `pytest --cov=...` (via `pytest-cov`) raises
 `ImportError: cannot load module more than once per process` on numpy
-import, use `coverage run -m pytest` followed by `coverage report` instead
-— same result, different import path.
+import, use `coverage run -m pytest` followed by `coverage report` instead.
+Same result, different import path.
 
 The `sh scripts/install_hooks.sh` pre-push hook requires Git Bash to
 install, and does not fire when pushing from plain PowerShell. Run it via
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,6 +32,12 @@ git switch -c my-change         # never work on main
 pip install -e ".[dev]"          # editable install so the working tree is what's tested
 sh scripts/install_hooks.sh      # pre-push hook: runs the suite before every push
 ```
+> **Windows contributors:** `make` is not available in PowerShell by default,
+> so `make ci` and `make riskcov` below will not run as written. See
+> [Testing on Windows](#testing-on-windows) for the equivalent commands.
+> `sh scripts/install_hooks.sh` also requires Git Bash; the installed
+> pre-push hook does **not** fire when pushing from plain PowerShell, only
+> from a POSIX-compatible shell (confirmed by testing).
 
 Install editable: a non-editable copy in site-packages will shadow your edits
 under pytest and silently run stale code.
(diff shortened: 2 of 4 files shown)


shivamlalakiya (changes requested):
Thank you for this. It is the first documentation in the repository written by someone who actually ran the gate on Windows, and the parts that are right are right for the reason that matters: they were tested, not guessed. The pre-push-hook finding in particular is one I could not have written.

One blocking change, then this goes in.

### The coverage floor must not be hardcoded

```powershell
python -m coverage report --fail-under=92
```

`92` is defined once, in `pyproject.toml` under `[tool.coverage.report]` (`fail_under = 92`), and `AGENTS.md` says explicitly not to repeat it anywhere else: "Do not hardcode either number anywhere else; that drift is what issues #21 and #22 exist to fix." A contributor who raises the floor in `pyproject.toml` will not think to grep `CONTRIBUTING.md`, and Windows contributors then run a gate one number looser than CI's without any signal that they are.

The fix is to delete the flag. `coverage report` reads `fail_under` from `pyproject.toml` on its own, so the bare command enforces exactly what `make coverage` enforces:

```powershell
python -m coverage report
```

Worth noting that your own risk-tier paragraph already gets this right by pointing at the `Makefile` variables rather than copying their values. The same instinct, applied one paragraph earlier, is all this needs.

### Two smaller things, both optional

- `make ci` runs `coverage`, which depends on `check`, so a collection-error guard runs before the suite does. The PowerShell list skips it. Adding `python -m pytest tests/ --collect-only -q` as the first line keeps the two paths equivalent.
- The CHANGELOG entry leaves trailing whitespace on the blank line after `Closes #195.`

### The 11 failures you found are a real bug, and yours

Please do not treat that as out of scope work you have to carry. I confirmed it: `ensure_local_path` calls `urlparse(str(path)).scheme` and checks it against `("", "file")`, and `urlparse(r"C:\data\gifts.csv").scheme` is `'c'`, so every absolute Windows path is rejected by every public entry point that takes one. That is a functional bug on Windows, not a test artifact. I have filed it as #217 with you credited for the report. Take it if you want it; leave it if you would rather not.

shivamlalakiya:
Rather than leave this waiting on a round trip, I pushed the change myself in c8a6e5c and merged `main` in. My earlier review stands dismissed by that commit; treat it as resolved.

What changed on top of your branch:

- `python -m coverage report --fail-under=92` is now `python -m coverage report`. Verified rather than assumed: `python -m coverage debug config` reports `fail_under: 92.0` read from `pyproject.toml`, so the bare command enforces the same floor CI does, and the number stays defined in one place. Added two sentences under the block saying so, since the next person to read that section will otherwise wonder where the gate went.
- Added `python -m pytest tests/ --collect-only -q` to the list. `make ci` runs `coverage`, which depends on `check`, so the collection guard runs on the Unix path; without it the PowerShell list was nearly equivalent rather than equivalent.
- Removed the two em dashes the branch introduced, one in `CONTRIBUTING.md` and one in your `CONTRIBUTORS.md` entry. Not a style nit I would normally spend your time on: #91 cleared every em dash out of this repository's prose and `main` has none, so a new one is a regression. Worth knowing for your next PR here.
- Stripped the trailing whitespace on the blank line in the CHANGELOG entry.

Everything else is yours as written. The Windows content itself is the whole value of this PR and I did not touch it.

The bug your testing turned up is #217, filed with you credited. It is real and it is not small: `ensure_local_path` rejects every absolute Windows path because `urlparse` reads the drive letter as a URI scheme, which breaks four public entry points on Windows. It is labelled `good first issue` and unclaimed if you want it.

Thank you for this. Documentation written by someone who actually ran the commands on the platform is worth several times documentation written by someone imagining them.

PR #216, opened 2026-09-10
test: exercise MovesManagementClassifier feature_names_in_ (#200)

### LINKED ISSUE #200: test: `MovesManagementClassifier.fit` is never given a DataFrame, so its `feature_names_in_` assignment is unexercised
### What's wrong

`MovesManagementClassifier.fit` sets `feature_names_in_` itself when the fitted `X`
carries column names, but no test in the repository ever fits it on a DataFrame. The
`stage_Xy` fixture in `tests/test_moves.py` builds `X` with
`np.random.default_rng(0).random((30, 5))`, `tests/test_sklearn_compliance.py` runs
`check_estimator` on arrays, and the class is only constructed with parameters in
`tests/test_public_api_contract.py`. A branch-coverage run reports the assignment line
as missed.

The consequence is not cosmetic. Every other public estimator in this project is
expected to record the training column names, and `sklearn`'s own
`get_feature_names_out` plumbing reads that attribute. Right now nothing would fail if
the assignment were deleted, mistyped, or reordered after `n_features_in_`, and the
same class was already the subject of a real defect of exactly this shape in
`CRMCleaner` (#157): an attribute that only exists after a named fit, read without a
guard.

### Where

| File | Locate with |
|---|---|
| `philanthropy/models/_moves.py` | `grep -n "feature_names_in_" philanthropy/models/_moves.py` |
| `tests/test_moves.py` | `grep -n "def stage_Xy" tests/test_moves.py` |

### What to change

No source change. This is test-only.

1. Read the `stage_Xy` fixture and `test_predict_proba_has_one_column_per_stage` in
   `tests/test_moves.py` for the fixture and assertion style used there.
2. Add a test that wraps the same `X` in a `pandas.DataFrame` with named columns, fits
   the classifier on it, and asserts:
   - `list(clf.feature_names_in_)` equals the DataFrame's column list, in order;
   - `clf.n_features_in_` equals the column count;
   - `predict` on the same DataFrame returns one label per row, so the named path is
     exercised end to end rather than only at fit time.
3. Add a one-line assertion (or a second test) that after an array fit the estimator
   has no `feature_names_in_`, which is what `sklearn` specifies and what pins the
   `hasattr` guard rather than the happy path alone.

Do not add a `feature_names_in_` fallback to this class as part of the same PR. If you
think one is warranted, say so in the PR description and leave it for a separate issue;
this one is about the missing test.

### Tests to add or extend

- File: `tests/test_moves.py`
- Add: `test_dataframe_fit_records_feature_names`
- Add: `test_array_fit_records_no_feature_names`

### Done when

```bash
# Exits 0 only once the named-fit line runs. Run it before opening the PR.
python -m pytest tests/test_moves.py \
  --cov=philanthropy.models._moves --cov-report=term-missing -q
make ci
```

`philanthropy/models/_moves.py` reports no missed lines in that report.

### First time here?

Read [CONTRIBUTING.md](../../CONTRIBUTING.md) and [AGENTS.md](../../AGENTS.md).
Add a `## [Unreleased]` CHANGELOG entry and yourself to `CONTRIBUTORS.md` in the
same PR.

HarshRajSinghania:
## Summary
Adds the missing `MovesManagementClassifier` tests requested in #200 so `feature_names_in_` is exercised on a named DataFrame fit and confirmed absent after an array fit.

## Motivation
`fit` assigns `feature_names_in_` when `X` has columns, but the suite only fitted arrays (plus a later partial DataFrame check that did not cover `n_features_in_`, `predict`, or the array-fit negative path). Nothing would fail if that assignment were deleted.

## Implementation
Test-only. In `tests/test_moves.py`:
- `test_dataframe_fit_records_feature_names` wraps the existing `stage_Xy` fixture in a DataFrame, asserts column names and `n_features_in_`, and calls `predict` on the same frame.
- `test_array_fit_records_no_feature_names` asserts the estimator has no `feature_names_in_` after an array fit.

No production-code change. CHANGELOG `[Unreleased]` and CONTRIBUTORS updated per project guidelines.

## Testing
Could not run the full suite in this environment: PyPI/sklearn install failed with repeated 502s from the package index proxy. Checks that *were* done:
- Reviewed `philanthropy/models/_moves.py` `fit` (`hasattr(X, "columns")` assignment and `n_features_in_`).
- Confirmed the new tests match the names and assertions specified in #200.
- Existing tests in the file were left unchanged.

Please run on CI:
```bash
python -m pytest tests/test_moves.py --cov=philanthropy.models._moves --cov-report=term-missing -q
make ci
```

Closes #200.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 ## [Unreleased]
 
 ### Added
+- Tests for `MovesManagementClassifier` now fit a named DataFrame and an
+  array so `feature_names_in_` is both recorded and absent on the paths
+  sklearn specifies. Closes #200.
 - `EncounterRecencyTransformer` gains an `as_of` parameter, the last dated
   transformer without one. Encounters dated after it are blanked to `NaT`
   before the features are computed, so a clinical encounter that had not
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -71,6 +71,11 @@ contribution. Code, docs, tests, and review all count.
   the useful invalid-timezone error (closes
   [#201](https://github.com/PhilanthroPy-Project/PhilanthroPy/issues/201)).
 
+- **Harsh Raj Singhania** ([@HarshRajSinghania](https://github.com/HarshRajSinghania)):
+  added DataFrame and array-fit tests so `MovesManagementClassifier.feature_names_in_`
+  is exercised rather than only assigned (closes
+  [#200](https://github.com/PhilanthroPy-Project/PhilanthroPy/issues/200)).
+
 ## Getting listed
 
 Add yourself here in the same pull request as your change: one line, your name
--- a/tests/test_moves.py
+++ b/tests/test_moves.py
@@ -58,19 +58,22 @@ def test_action_priority_summary_counts_every_donor(stage_Xy):
     assert set(summary) <= set(_STAGES)
 
 
-def test_fit_with_dataframe_sets_feature_names_in():
-    X = pd.DataFrame(
-        np.random.default_rng(0).random((30, 5)),
-        columns=["age", "income", "donations", "engagement", "years_active"],
-    )
-    y = np.asarray(_STAGES * 10)
+def test_dataframe_fit_records_feature_names(stage_Xy):
+    X_arr, y = stage_Xy
+    columns = ["age", "income", "donations", "engagement", "years_active"]
+    X = pd.DataFrame(X_arr, columns=columns)
 
     clf = MovesManagementClassifier(max_iter=10, random_state=0).fit(X, y)
 
-    np.testing.assert_array_equal(
-        clf.feature_names_in_,
-        np.array(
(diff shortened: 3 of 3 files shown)


shivamlalakiya:
Merged. Thank you.

Two notes for the record.

You were right that #200 was partly stale when I filed it. `e8914ca` (2026-08-24, @stoppo22) already fitted a DataFrame, so the assignment was not entirely unexercised and the issue's "no test in the repository ever fits it on a DataFrame" was wrong. Replacing that test rather than adding beside it was the right call: it keeps the old assertion and adds the three things it missed (`n_features_in_`, `predict` on the named path, and the array-fit negative case that pins the `hasattr` guard). Saying so in the PR description instead of quietly working around it is the part I appreciate most.

Not being able to install the dependencies is fine, and naming the commands you could not run was the right way to handle it. GitHub was also reporting the branch as conflicting with `main`, which was real rather than stale: this repository sets `CHANGELOG.md merge=union` in `.gitattributes`, so a local `git merge` resolves two bullets landing in the same place silently, while GitHub, which does not read that file, sees a conflict. I merged `main` into your branch and pushed it here, which cleared it. All 14 checks passed on the merged head before merge.

PR #218, opened 2026-09-10
fix(validation): accept Windows drive paths

### LINKED ISSUE #217: bug: `ensure_local_path` rejects every absolute Windows path; `urlparse` reads the drive letter as a scheme
Reported by @AsavariCharati while verifying #215 on Windows: 11 failures in the local suite, all from the same cause. Confirmed here.

### What's wrong

`ensure_local_path` decides whether a path is local by reading `urlparse(str(path)).scheme` and comparing it against `_LOCAL_SCHEMES = ("", "file")`. On Windows an absolute path begins with a drive letter and a colon, which `urlparse` parses as a URI scheme:

```python
>>> from urllib.parse import urlparse
>>> urlparse(r"C:\data\gifts.csv").scheme
'c'
```

`'c'` is not in `_LOCAL_SCHEMES`, so the call raises:

```
ValueError: `data` must be a local file path (no network reads), got scheme 'c' in 'C:\\data\\gifts.csv'.
```

Every absolute Windows path fails, for every drive letter. This is not a test-only problem. The guard sits on four public entry points, so on Windows the package rejects the ordinary way of naming a file:

| Caller | Parameter |
|---|---|
| `philanthropy/cli.py:66` | `data` |
| `philanthropy/cli.py:210` | `args.data` |
| `philanthropy/preprocessing/_grateful_patient.py:235` | `encounter_path` |
| `philanthropy/preprocessing/_encounters.py:435` | `encounter_path` |

Relative paths (`data\gifts.csv`) and forward-slash absolute paths without a drive (`/data/gifts.csv`) still work, which is why CI has never seen this: the Windows job does not exist, and every test fixture uses `tmp_path`-relative or POSIX-style paths.

### Where

| File | Locate with |
|---|---|
| `philanthropy/utils/_validation.py` | `grep -n "_LOCAL_SCHEMES\|urlparse" philanthropy/utils/_validation.py` |
| `tests/test_no_network.py` | `grep -n "rejects_remote_scheme" tests/test_no_network.py` |

### What to change

In `ensure_local_path`, treat a single-character scheme as a Windows drive letter rather than as a URI scheme:

```python
scheme = urlparse(str(path)).scheme
if len(scheme) == 1 and scheme.isalpha():   # Windows drive letter, not a URI scheme
    scheme = ""
```

Do this rather than platform-branching on `os.name`. A `C:\...` path handed to a Linux process is still not a network read, and the check should not give a different answer depending on where it runs. The security guarantee this function exists for is unaffected: every scheme it is written to reject (`http`, `https`, `s3`, `gs`, `ftp`, `gcs`, `abfs`) is at least two characters, and no single-letter scheme that pandas would follow to the network exists.

Keep the rejection message and the `ValueError` type exactly as they are. `tests/test_no_network.py` matches on `"must be a local file path"` and #123 pinned this behaviour.

### Tests to add or extend

- File: `tests/test_no_network.py`, alongside the existing `test_encounter_path_rejects_remote_scheme` / `test_cli_data_path_rejects_remote_scheme`
- Add `test_windows_drive_letter_path_is_local`: `ensure_local_path(r"C:\data\gifts.csv")` returns the path unchanged, and asserts the same for a lowercase drive letter. These must pass on Linux and macOS too, since the fix is not platform-branched.
- Add a negative case in the same test file so the loosening is pinned as narrow: a two-character scheme such as `s3://bucket/gifts.csv` still raises.

### Done when

```bash
python -m pytest tests/test_no_network.py -q
python -m pytest tests/ --cov=philanthropy.utils._validation --cov-report=term-missing -q
make ci
```

On a Windows machine, the 11 failures @AsavariCharati reported are gone. If you do not have Windows, the tests above are still decisive: they exercise the same code path on any platform, because the parsing bug is in `urlparse`, not in the filesystem.

### Note on Windows CI

This slipped through because the test matrix is `ubuntu-latest` and `macos-latest` only. Adding a `windows-latest` job is the durable fix and is worth its own issue; it is deliberately not part of this one, which should stay a small, reviewable bug fix.

### First time here?

Read [CONTRIBUTING.md](../blob/main/CONTRIBUTING.md) and [AGENTS.md](../blob/main/AGENTS.md). Add a `## [Unreleased]` CHANGELOG entry and yourself to `CONTRIBUTORS.md` in the same PR.

laichouchi:
## Summary
- treat one-character URL schemes as Windows drive letters in ensure_local_path
- keep rejecting real network schemes such as https, s3, gs, and gcs
- add regression coverage for uppercase and lowercase drive letters
- document the fix and add the contributor entry

Closes #217

## Validation
- make ci
- make riskcov
- 2,070 tests passed, 29 skipped
- total coverage: 98.22%
- risk-tier coverage: 98%

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,8 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
   runnable `as_of` walkthrough of the same failure. Closes #209.
 
 ### Fixed
+- `ensure_local_path` now accepts absolute Windows drive-letter paths such as
+  `C:\\data\\gifts.csv` without weakening rejection of network URLs. Closes #217.
 - `EncounterRecencyTransformer` no longer catches timezone conversion errors
   and retries with `tz_localize` after parsing with `utc=True`. The retry was
   unreachable for valid timezone inputs and replaced the useful
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -10,6 +10,9 @@ contribution. Code, docs, tests, and review all count.
 
 ## Contributors
 
+- **Leyn.cx** ([@laichouchi](https://github.com/laichouchi)): fixed Windows
+  drive-letter paths being rejected as URI schemes (closes #217).
+
 - **Chris Chen** ([@fuleinist](https://github.com/fuleinist)): replaced a
   tautological doctest in `FiscalYearGroupedSplitter` with one that can actually
   fail, and added the missing no-leakage test for the default `gap_years=0`
--- a/philanthropy/utils/_validation.py
+++ b/philanthropy/utils/_validation.py
@@ -72,6 +72,11 @@ def ensure_local_path(path: _PathT, param_name: str = "path") -> _PathT:
     from urllib.parse import urlparse
 
     scheme = urlparse(str(path)).scheme
+    # urlparse treats the drive letter in an absolute Windows path such as
+    # C:\\data\\gifts.csv as a one-character URI scheme. It is still a local
+    # path, even when this check runs on a non-Windows host.
+    if len(scheme) == 1 and scheme.isalpha():
+        scheme = ""
     if scheme not in _LOCAL_SCHEMES:
         raise ValueError(
             f"`{param_name}` must be a local file path (no network reads), "
--- a/tests/test_no_network.py
+++ b/tests/test_no_network.py
@@ -135,6 +135,13 @@ def test_cli_data_path_rejects_remote_scheme(no_network):
         _read_csv("gs://bucket/prospects.csv")
 
 
+def test_windows_drive_letter_path_is_local():
+    from philanthropy.utils._validation import ensure_local_path
+
+    for path in (r"C:\data\gifts.csv", r"z:\data\gifts.csv"):
+        assert ensure_local_path(path) == path
+
+
 def test_local_paths_still_load_under_the_guard(no_network, tmp_path):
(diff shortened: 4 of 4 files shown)


shivamlalakiya:
Reviewed and merging. The fix is in the right place: `ensure_local_path` is the single chokepoint all four public entry points route through (`cli._read_csv`, `cli.main`, `EncounterTransformer.fit`, `GratefulPatientFeaturizer.fit`), so one guard there fixes every one of them rather than four separate patches.

Checked the thing that matters about the widened predicate: no scheme `pandas` would follow over the network is one character. Every protocol `fsspec` registers is two or more (`s3`, `gs`, `gcs`, `http`, `https`, `ftp`, `sftp`, `abfs`, `hdfs`, `memory`), so `len(scheme) == 1` cannot collide with a real remote read. The rejection path keeps working, and `test_encounter_path_rejects_remote_scheme` and `test_cli_data_path_rejects_remote_scheme` still cover it.

Two details worth noting for the record, both already handled by your test: `urlparse` lowercases the scheme, so `C:` and `c:` both arrive as `'c'`, and the test pins uppercase and lowercase separately anyway. And the guard is host-independent by design, which is what the comment says: it has to run identically on Linux CI, or the coverage of this branch would be unreachable in the matrix.

Full matrix green on py3.9 through py3.13 plus macOS, so this is verified on the platforms CI can reach. Thank you for turning a report you found in someone else's PR into the fix, and for the changelog and contributor entries without being asked twice.

Issue #55, opened 2026-08-11
test: `FiscalYearGroupedSplitter.split` never exercises the no-prior-training-data skip branch


shivamlalakiya:
### What's wrong

`FiscalYearGroupedSplitter.split` skips a fold with `continue` when no training
row exists strictly before `test_fy - gap_years`. Every fixture in
`tests/test_model_selection.py` has enough fiscal years of history that this
never fires, so the skip path is untested — a splitter silently dropping a fold
short of `n_splits` is exactly the kind of thing that should be pinned with a
regression test.

### Where

| File | Locate with |
|---|---|
| `philanthropy/model_selection/_temporal_donor_splitter.py` | `grep -n "No training data before this test year" philanthropy/model_selection/_temporal_donor_splitter.py` |

### What to change

No source change — this is test-only.

1. Open `tests/test_model_selection.py` and look at the existing `split()`
   fixtures (locate with `grep -n "def test_" tests/test_model_selection.py`).
2. Add a test with fiscal years arranged so the *earliest* requested test year
   has no fiscal year strictly before it once `gap_years` is applied (e.g.
   `groups` starting at the very first fiscal year present, with `n_splits`
   covering it). Assert the splitter yields fewer than `n_splits` folds instead
   of raising or yielding an empty training set.

### Tests to add or extend

- File: `tests/test_model_selection.py`
- Add: `test_split_skips_fold_with_no_prior_training_data`

### Done when

```bash
python -m pytest tests/test_model_selection.py --cov=philanthropy.model_selection --cov-report=term-missing -q
```

`philanthropy/model_selection/_temporal_donor_splitter.py` reads 100% covered
with an empty `Missing` column.

jimoholayinka01:
Hi! I'd like to work on this issue. I can add the regression test for the no prior training data skip branch and verify the splitter coverage as described. Thanks!

Issue #53, opened 2026-08-11
test: cli's `_score_array` never exercises the plain `predict_proba` fallback branch


shivamlalakiya:
### What's wrong

`_score_array` picks `predict_affinity_score` when the model has one, otherwise falls
back to `model.predict_proba(X)[:, 1]`. Every `score` invocation in
`tests/test_cli.py` scores whatever `train` just produced with the **default**
`--model` (`DonorPropensityModel`), which has `predict_affinity_score` — so the
fallback line never runs. Two of the four CLI-documented models,
`LapsePredictor` (`predict_lapse_score`, no `predict_affinity_score`) and
`PlannedGivingIntentScorer` (`predict_intent_score`, no `predict_affinity_score`),
would hit that fallback, but nothing ever trains-then-scores either of them.

### Where

| File | Locate with |
|---|---|
| `philanthropy/cli.py` | `grep -n "_score_array" -A 3 philanthropy/cli.py` |

### What to change

No source change — this is test-only.

1. Open `tests/test_cli.py` and look at `test_cli_train_accepts_every_documented_model`
   (locate with `grep -n "def test_cli_train_accepts_every_documented_model" tests/test_cli.py`)
   for the existing `--model` parametrization style.
2. Add a new test that runs `train` with `--model PlannedGivingIntentScorer` (or
   `LapsePredictor`), then `score` against the same data, and asserts the `score`
   column exists and has no `NaN` values — proving the `predict_proba` fallback path
   ran end to end without an `AttributeError`.

### Tests to add or extend

- File: `tests/test_cli.py`
- Add: `test_cli_score_falls_back_to_predict_proba_without_affinity_score`

### Done when

```bash
python -m pytest tests/test_cli.py --cov=philanthropy.cli --cov-report=term-missing -q
```

`philanthropy/cli.py` reads 100% covered with an empty `Missing` column.

jimoholayinka01:
Hi! I'd like to work on this issue. I can add the regression test for the predict proba fallback path and verify the CLI coverage as described. Thanks!

Issue #50, opened 2026-08-11
test: PropensityScorer.predict_proba never exercises its single-class branch


shivamlalakiya:
### What's wrong

`PropensityScorer.predict` and `PropensityScorer.predict_proba` each have a short-circuit for the case where `fit` only ever saw one class (`len(self.classes_) == 1`). The one in `predict` is exercised today — scikit-learn's own `check_classifiers_one_label` conformance check (run in `tests/test_sklearn_compliance.py`) fits the estimator on single-label data and calls `predict`. That check does not call `predict_proba`, so `predict_proba`'s parallel branch — returning a `(n_samples, 1)` column of ones, exactly as its own docstring's `Returns` section promises — has no test at all.

### Where

| File | Locate with |
|---|---|
| `philanthropy/models/_propensity_baseline.py` | `grep -n "def predict_proba" -A 6 philanthropy/models/_propensity_baseline.py` |

The match on `classes_) == 1` a few lines below `def predict_proba` is the one this issue is about. The identical-looking guard inside `def predict` (above it in the same file) already has coverage — leave that one alone.

### What to change

No source change — this is test-only.

1. Open `tests/test_propensity.py` and look at `test_propensity_scorer_predict_proba_shape` (locate with `grep -n "def test_propensity_scorer_predict_proba_shape" tests/test_propensity.py`) for the existing fixture/assertion style.
2. Add a new test after it that fits `PropensityScorer` on a `y` containing only one label, then calls `predict_proba` and checks the shape and values match the docstring's promise.

### Tests to add or extend

- File: `tests/test_propensity.py`
- Add: `test_propensity_scorer_predict_proba_single_class`

```python
def test_propensity_scorer_predict_proba_single_class():
    X = np.ones((5, 3))
    y = np.zeros(5)
    clf = PropensityScorer().fit(X, y)
    proba = clf.predict_proba(X)
    assert proba.shape == (5, 1)
    assert np.all(proba == 1.0)
```

### Done when

```bash
python -m pytest tests/test_propensity.py --cov=philanthropy.models._propensity_baseline -q
python -m coverage report --include='philanthropy/models/_propensity_baseline.py' --fail-under=100
make ci
make riskcov
```

The `coverage report --fail-under=100` line exits non-zero until `_propensity_baseline.py` reads fully covered; `make ci` and `make riskcov` must also stay green.

### First time here?

[...]

PR #222, opened 2026-09-14
test: pin invalid timezone error on EncounterRecencyTransformer

### LINKED ISSUE #204: test: no test pins the invalid-timezone error that #202 restored in `EncounterRecencyTransformer`
### What's wrong

#202 (closing #201) deleted a `try`/`except` in `EncounterRecencyTransformer._parse_dates` that caught every exception from `.dt.tz_convert(self.timezone)` and retried with `.dt.tz_localize(self.timezone)`. The removal is correct: the same method parses with `utc=(self.timezone is not None)`, so whenever the fallback could run the series is already tz-aware and `tz_convert` cannot fail for a valid timezone. The only way the `except` ever fired was an **invalid** timezone name, and there it replaced the lookup error with a `TypeError` reading "Already tz-aware", which names the wrong problem entirely.

No test covers that. `tests/test_encounter_timezone.py` has five tests and all five pass a valid timezone or none at all; there is no assertion anywhere that an invalid timezone raises, and none about which exception comes out. So the behaviour #202 restored is unguarded: someone could reintroduce a bare `except` around that call, or swap the `utc=` argument back to a constant, and the whole suite would stay green while the misleading error came back.

This is the same shape as #200 and #157: a behaviour that only exists on a path no test walks.

### Where

| File | Locate with |
|---|---|
| `philanthropy/preprocessing/_encounter_recency.py` | `grep -n "tz_convert" philanthropy/preprocessing/_encounter_recency.py` |
| `tests/test_encounter_timezone.py` | `grep -n "def test_encounter_recency_localises" tests/test_encounter_timezone.py` |

### Expected behaviour, so it does not have to be guessed

An invalid timezone name must surface as a **lookup error naming the offending timezone**, not as a complaint about the data.

The concrete type differs by pandas major, so do not pin it directly. Both of these were run against `main` at the merge of #202:

| Environment | Exception | `str(e)` |
|---|---|---|
| pandas 2.3.3 (pytz present) | `pytz.exceptions.UnknownTimeZoneError` | `"'Not/AZone'"` |
| pandas 3.0.5 (resolves via `zoneinfo`) | `zoneinfo.ZoneInfoNotFoundError` | `"'No time zone found with key Not/AZone'"` |

`pyproject.toml` declares `pandas>=2.0` with no upper bound and the CI matrix spans py3.9 to py3.13 plus a lowest-direct floors job, so **both rows really happen in CI**. Pinning `pytz.exceptions.UnknownTimeZoneError` would fail on the pandas 3 jobs, and `pytz` is not a declared dependency of this project, so importing it in a test is not safe either.

**What to pin instead:** both exceptions subclass `KeyError`, and the pre-#202 bug raised `TypeError("Already tz-aware")`, which does not. So `KeyError` plus the timezone name in the message discriminates the regression on every supported pandas, with no `pytz` import:

```python
with pytest.raises(KeyError, match="Not/AZone"):
    ...
```

Do not match on `Exception` or `ValueError`. Both would pass against the bug, which is the thing this test exists to catch.

### What to change

No source change. This is test-only.

1. Read `test_encounter_recency_localises_to_a_non_utc_timezone` in `tests/test_encounter_timezone.py` for the fixture and assertion style used there; reuse its two-row `X`.
2. Add a test that constructs the transformer with a timezone name that does not exist, calls `fit_transform`, and asserts with `pytest.raises(KeyError, match=...)` that the offending name appears in the message.
3. Add an assertion (in the same test or a second one) that the message does **not** contain "Already tz-aware", so the test fails loudly if the masking fallback is ever reintroduced rather than silently passing on the wrong exception.

`tests/conftest.py` pins `TZ=UTC`, so this is reproducible off a CI runner with no extra setup.

Please keep the PR to the test. If you think `_parse_dates` should validate `timezone` earlier, or re-raise something friendlier, say so in the PR description and leave it for a separate issue; `__init__` doing no validation is a deliberate scikit-learn convention here (see AGENTS.md) and changing it is a larger decision than this issue.

### Verifying

```
python -m pytest tests/test_encounter_timezone.py -q
make ci
```

Every PR also needs a `## [Unreleased]` entry in `CHANGELOG.md` and your own name added to `CONTRIBUTORS.md`, in the same PR. CI enforces both.

Follow-up to #201 / #202.

HarshRajSinghania:
Closes #204.

## Summary
Adds a regression test so an invalid `timezone=` on `EncounterRecencyTransformer` raises `KeyError` that names the zone. The pre-#202 fallback turned a bad zone into `TypeError("Already tz-aware")`; this test would fail if that path returned.

No production code change. `__init__` still does not validate `timezone` (sklearn convention noted in the issue).

## Motivation
#204: `tests/test_encounter_timezone.py` only used valid timezones or none. The behaviour restored in #202 was unguarded.

## Implementation
- Reuse the two-row fixture from `test_encounter_recency_localises_to_a_non_utc_timezone`.
- Construct the transformer with `timezone="Not/AZone"` and call `fit_transform`.
- Assert `pytest.raises(KeyError, match="Not/AZone")` so both pandas 2 (pytz) and pandas 3 (zoneinfo) are covered without importing pytz.
- Assert the message does not contain `Already tz-aware`.
- `CHANGELOG.md` `[Unreleased]` entry and `CONTRIBUTORS.md` line, as required by the project.

## Testing
Intended local commands from the issue:
```
python -m pytest tests/test_encounter_timezone.py -q
make ci
```
This environment could not install scikit-learn (PyPI 502), so the new test was not executed here. CI on this PR should run the file and the full matrix.


shivamlalakiya:
Closing this as a duplicate: #204 was claimed by @Kaap10 on 2026-09-12 and confirmed on the issue the same day, and #221 has now merged as dba8cd5.

That is the claim-before-PR rule from #187 working as intended, and it is on me that it is easy to miss: the claim is a comment on the issue rather than an assignee, so the issue looks unclaimed from the list view. Checking the issue thread before starting is the only reliable way right now.

Nothing wrong with the test itself. If you want another, #50, #53 and #55 are all open with zero comments, so no claim to collide with:

- #50 `PropensityScorer.predict_proba` single-class branch
- #53 cli `_score_array` plain `predict_proba` fallback branch
- #55 `FiscalYearGroupedSplitter.split` no-prior-training-data skip branch

Thanks for #216 and for this one.

